Issue #5, opened 2019-07-10
Refactor Code


TimonGaebelein:
Prepare Code for Touchscreen Input

PR #6, opened 2019-07-10
Refactor

TimonGaebelein:
Refactor the code into single components to make it easier to create the touchscreen interface. This should also enable me to do both (touchscreen and non-touchscreen) via a config flag.

Just needs some final testing before merging


TimonGaebelein:
Seems to work. Closes #5.

Issue #85, opened 2019-08-13
Design an App Icon


UnchartedBull:
General Guidelines:
- should be a png with transparent background
- 512x512px
- I would like it to reflect the connection to OctoPrint but also look modern, but I'm open for anything.
- ColorPalette used for OctoDash is this one: https://flatuicolors.com/palette/gb so probably a good point to start
- It should look modern and follow the general design of OctoDash

hessius:
I saw that someone was helping out with this on reddit, but if that's stalled it would be good to get some specifications (sizes, format) and if there are any guidelines (palette, general ideas etc)

UnchartedBull:
Yes it went silent, unfortunately. I may ask that guy for an update but I don't want to push anyone. I added some information to the issue.

@hessius thanks for all your feedback and suggestions so far. If you would like to become a contributer and help me out in some places let me know. I'll be happy to add you.

hessius:
Unfortunately I'm not familiar with angular, so I think "onboarding" me wouldn't be worth the hassle.

I really like the work you've done, so I'm happy to help - Don't worry though, I'm out of ideas now and won't keep submitting issues at this pace ;) 

I do have an idea for an icon though and will se if I can get it to an okay state :)

hessius:
So here's my take, I wanted something simple that could also be incorporated into the ui if needed / wanted, I'm a fan using single coloured logos as they can be reproduced on any background with any background and colour scheme.

So here, presented in white on the OctoDash background colour is simply a combination of a dashboard and an octopus - OctoDash!


![octodash-icon](https://user-images.githubusercontent.com/1499030/66839189-54ce0000-ef66-11e9-8b56-c67a9c4c15da.png)


_Note, while I've made minor adjustments, this is made out of icons from the noun project which can be used even commercially but without a paid license (a one time fee of a few dollars) require attribution (shouldn't be a problem but wanted to mention it)_

UnchartedBull:
That icon does look quite cool. Would you mind adding some colors (like the green tone from the OctoPrint logo)? I've tried to sprinkle color into OctoDash where it was possible to make it look a little bit friendlier, so maybe a few spots of colors does no harm to the logo.

I also used the noun project for almost all other icons, so don't worry, they will get their mentioning soon (currently only in the icons.txt file)

hessius:
Yeah absolutely, I can get on that tonight. Do you want the same colour everywhere?
[...]

PR #242, opened 2019-10-22
Add svg icons

hessius:
As discussed in #85 here are svg versions of the icon selected


README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
-    <img width="200" height="200" src="https://github.com/UnchartedBull/OctoDash/raw/master/build/icon.png">
-    <br>
+    <img width="300" height="300" src="./src/assets/icon/icon-main-dark-title.svg">
```
UnchartedBull:
```suggestion
    <img width="300" height="300" src="https://github.com/UnchartedBull/OctoDash/raw/master/src/assets/icon/icon-main-dark-title.svg">
```

GitHub doesn't resolve paths AFAIK, therefore you need to include the whole URL.

hessius: @UnchartedBull Have you tried this? When I tried with an absolute path it didn't resolve when using an svg-file even when specifying the raw output.
But if it works for you I'll of course resolve it!

hessius: It does however resolve the relative path, you can see it in action here:

https://github.com/hessius/OctoDash

UnchartedBull: That's funny. Maybe it does work because it is HTML and not Markdown. Pretty sure this won't work with Markdown. Then this is obsolete of course.

PR #503, opened 2020-03-20
Add temperature controls to main menu

harleyg321:
Set Hotend Temperature
**Features Added**
This allows hotend and heatbed temperatures to be set from the main menu by clicking on their status icon/temperature.

While this could be managed from the control menu, I like having the preheat and cooldown buttons on there to set both hotend and heatbed temperatures. This allows control of each individually from the main menu, for example to set just the hotend temperature to remove the filament without also heating up the bed.

**Potential Improvements**
To save having to increment up from 0, I've set a default starting value of 200 for the hotend and 60 for the heatbed. This could be something retrieved from config to let users set their own default value.

This only adds the ability to adjust temperatures. It would make sense to also allow fan speed control in the same way. However I couldn't see an easy way already implemented to set fan speed so I've added temperature controls for now.

### DIFF
--- a/src/app/printer-status/printer-status.component.html
+++ b/src/app/printer-status/printer-status.component.html
@@ -1,7 +1,9 @@
 <table class="printer-status" *ngIf="printerStatus">
     <tr>
         <td class="printer-status__value printer-status__value-1"
-            [ngClass]="{'printer-status__no-set-value': printerStatus.nozzle.set === 0}">
+            [ngClass]="{'printer-status__no-set-value': printerStatus.nozzle.set === 0}"
+            (click)="showQuickControlHotend()"
+            matRipple [matRippleUnbounded]="false">
             <img src="assets/nozzle.svg">
             <span class="printer-status__actual-value">{{ printerStatus.nozzle.current }}<span
                     class="printer-status__unit">°C</span></span>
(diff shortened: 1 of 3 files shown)


UnchartedBull:
Looks great. Code also looks fine on first inspection. I'll add the config part during the next week and will merge after that. Thank you!

harleyg321:
The ability to set fan speed probably isn't that useful, but for completeness I've also added it. I've used the printer status executeGCode function to issue an M106 Sxxx command to set the speed.

UnchartedBull:
I somehow messed up this PR. I'm new to accepting PR's from other contributors, so bear with me. I somehow created a new branch by accident (https://github.com/UnchartedBull/OctoDash/tree/harleyg321-feature/temp_control) to which I pushed my proposed changes. Please have a look there and let me know, whether you're fine with them. I will also add the config part to this branch as well.

Let men know, whether we should merge this branch to your repo first, or whether we should directly merge into master from there.

harleyg321:
Changes look good, I guess merging from that feature branch directly into master is probably easiest.

UnchartedBull:
Ok then this will be merged in #523. Please have a look at the latest changes, once you're happy just comment on the other PR, after that I'll merge it then :)

Issue #259, opened 2019-10-25
tune temperatures from status screen


mvoss96:
I think it would be a great feature to be able to tune the temperatures and the fan from the status screen.
When you touch f.e. the hotend icon it should show a slider where you can set the desired temperature.
Ideally the status screen should also show the current target temperature and not only the current temperature.

UnchartedBull:
OctoDash will show the target temperature if there is one. If the target temperature is 0 this will be omitted. You can also adjust the temperatures during print if you just tap the screen and go to the adjust menu.

I may consider adding something for adjusting if you're not printing in the future though ... I'm currently trying to postpone most of the new features so v2 will be ready and can be released finally.

harleyg321:
I hadn't actually seen this beforehand but I've just submitted a PR(#503) that would do this.

PR #523, opened 2020-03-27
Harleyg321 feature/temp control

UnchartedBull:


src/app/printer-status/printer-status.component.html
```diff
@@ -1,25 +1,74 @@
...
     </tr>
-</table>
\ No newline at end of file
+</table>
+<div class="quick-control" *ngIf="view !== QuickControlView.NONE" id="quickControl">
+    <table class="top-bar" (click)="hideQuickControl()">
+        <tr>
+            <td class="top-bar__back"><img src="assets/back.svg" class="top-bar__back-icon" matRipple
+                    [matRippleUnbounded]="false" />back</td>
+            <img [src]="'assets/' + (view === QuickControlView.HOTEND ? 'nozzle.svg' : view === QuickControlView.HEATBED ? 'heat-bed.svg' : 'fan.svg')"
```
harleyg321: Currently the icon doesn't display. Moving this down 2 rows so the image is outside the table fixes the issue.

Alternatively, changing the class to top-bar__center-icon and keeping it where it is inside the table row gets it to display although it then isn't centered.

src/app/printer-status/printer-status.component.scss
```diff
@@ -67,11 +67,98 @@
...
+        right: 2vw;
+        top: 3vh;
+    }
+
+    &__controller {
+        width: 35vw;
+        margin: 0 auto;
+        border: solid 0.6vw;
+        border-radius: 3vw;
+        margin-top: 5vh;
```
harleyg321:
```suggestion
```
Part of the changes you've made seem to have moved the controller box down the screen, removing the top margin gets it back to the center again.

src/app/printer-status/printer-status.component.html
```diff
@@ -1,25 +1,74 @@
...
+                class="quick-control__center-icon-small">
+        </tr>
+    </table>
+    <div class="quick-control__controller">
```
harleyg321:
```suggestion
            <td class="top-bar__back"><img src="assets/back.svg" class="top-bar__back-icon" matRipple
                    [matRippleUnbounded]="false" />back</td>
        </tr>
    </table>
    <img [src]="'assets/' + (view === QuickControlView.HOTEND ? 'nozzle.svg' : view === QuickControlView.HEATBED ? 'heat-bed.svg' : 'fan.svg')"
        class="quick-control__center-icon-small">
    <div class="quick-control__controller">
```

Currently the icon doesn't display. Moving rows 34/35 down 2 rows so the image is outside the table fixes the issue.

Alternatively, changing the class to top-bar__center-icon and keeping it where it is inside the table row gets it to display although it then isn't centered.

src/app/printer-status/printer-status.component.scss
harleyg321: Ah, it was moving the image inside the table that did it.

If you keep the image inside the table, I suggest making this change. If you move the image outside the table as in my other review comment, then this change should not be made.

UnchartedBull:
Somehow messed it up with the icon. I moved it into the table and everything is working for me now :) Will be released in 1.4.1 (should be in the next couple of days). And once again: Closes #259

PR #540, opened 2020-04-02
Go into standby if unable to contact OctoPrint

### LINKED ISSUE #539: Error always received on startup
**Describe the bug**
Every time I turn on my raspberry pi, I get a "Can't retrieve printer status" error before it goes into standby.

**Expected behavior**
Errors shouldn't be shown on startup while we are still waiting for OctoPrint to launch.

**General Information:**
 - OctoDash v1.4.1
 - OctoPrint v1.4.0
 - Klipper v0.8.0

**Additional context**
I think what is happening is OctoDash starts before OctoPrint has finished starting, so it is unable to contact OctoPrint, and displays an error. OctoPrint then starts up, at which point OctoDash is able to get the printer status and see that the serial connection has not been opened yet and goes into standby, but not before an error has already been shown (which persists onto the standby screen).

harleyg321:
If unable to connect to OctoPrint (signalled by an error status of 0 suggesting a network error, instead of a http error), go into standby.

Happy for you to close this PR if you would rather give OctoPrint some more time to finish launching when OctoDash starts, rather than going straight to the standby screen.

Fixes #539

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: https://paypal.me/TimonGaebelein
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,106 @@
-# Change Log
+# Changelog
+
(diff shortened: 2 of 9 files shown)


UnchartedBull:
I would rather like OctoPrint to show each error it may encounter. But having a Boot Grace Time might be a good thing. I may slightly refactor this later to only work for the first 60 seconds or so, so OctoPrint has enough time to boot up :)

UnchartedBull:
Ok I made the changes locally. Could you maybe check the checkbox “Allow edits from maintainers.”. This would allow me to push them into this PR :)

harleyg321:
Perhaps http requests could have a second attempt after something like 3-5 seconds later if an error is encountered, and only return the error if the second attempt is also not successful (atleast if the status is 0, because of a network error).

This would then work not only on boot, but also other situations when network errors are encountered (thinking someone may have octodash and octoprint on two devices, maybe the WiFi drops out intermittently), or atleast slow down the rate of new http error notifications shown.

harleyg321:
> Ok I made the changes locally. Could you maybe check the checkbox “Allow edits from maintainers.”. This would allow me to push them into this PR :)

It is already checked, not sure if there's anything else I need to do to allow you to commit.

UnchartedBull:
Mhh, I'll check again tomorrow to see what I did wrong. I basically did a 30 second boot grace time, which will surpress all 0 errors for the first 30 seconds.

UnchartedBull:
Since I didn't manage to do push into your branch I just created a PR for your branch. After that we can merge this into master :)

PR #542, opened 2020-04-02
Fix errors and unnecessary data shown when printing from SD

### LINKED ISSUE #541: Errors received when printing from sd
**Describe the bug**
When printing from klipper's virtual SD errors are constantly shown, alternating between "Can't retrieve Job Status" and "Can't load thumnail". Also OctoDash does not show the job progress screen.

**Expected behavior**
As I am using klipper which runs on the same raspberry pi as OctoPrint and OctoDash, if I am using my printer then OctoPrint and OctoDash are also running. Therefore information that is still available when printing from SD should be shown, while irrelevant data is hidden.

**General Information:**
 - OctoDash v1.4.1
 - OctoPrint v1.4.0
 - Klipper v0.8.0

**Additional context**
I have klipper's virutal sd directory set to the same as the OctoPrint upload folder. This way if I want to print from the printer directly, not via OctoPrint, the files are already available. This causes each file to show twice in the OctoPrint file list, the second one being the lowercase version of the uploaded filename. The file still only exists once in the uploads folder.

Therefore when printing one of the sdcard files the filename is not accessible through OctoPrint, but OctoDash still tries to get the thumbnail (and gets a 404 error). This causes a lot of error messages to show. The printer stats also becomes 'Printing from SD' which is not detected by OctoDash as a printing status, so the job progress screen does not show.

harleyg321:
When printing from SD, a thumbnail image should not be retrieved as the filename does not exist in OctoPrint.

Null values should be propagated through to the view instead of replaced with 0. This way values which are not available when printing from SD (listed below, or any other situation also resulting in it being null) can be hidden.

 - Filament used
- Time remaining, until OctoPrint has estimated a time remaining based on current print progress, otherwise it shows 0 remaining for the first minute or so of printing, which looks odd
- Layer progress, which shows as layer 0 of 0

Fixes #541

### DIFF
--- a/src/app/job-status/job-status.component.html
+++ b/src/app/job-status/job-status.component.html
@@ -5,9 +5,9 @@
         <div class="job-info__progress-percentage">{{ job.progress }}<span style="font-size: 40%">%</span></div>
     </div>
     <span class="job-info__filename">{{ job.filename }}</span> <br />
-    <span class="job-info__filament">{{ job.filamentAmount }}g Filament</span> <br />
-    <span class="job-info__time"><span
-            class="job-info__time-left">{{ job.timeLeft.value }}</span>{{ job.timeLeft.unit }} left,
+    <span class="job-info__filament" *ngIf="job.filamentAmount !== null">{{ job.filamentAmount }}g Filament</span> <br />
+    <span class="job-info__time"><span *ngIf="job.timeLeft.value !== null"><span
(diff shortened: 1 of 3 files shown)


UnchartedBull:
OctoDash doesn't support SD Card Files (since it's always a big hassle with the GCode information and stuff), that's why I haven't really checked those things. But this is a great compromise I think!

PR #546, opened 2020-04-05
added thumbnail of file being printed to print progress view

jneilliii:
Just a simple change to show the thumbnail on the right side of the print progress status.

image


UnchartedBull:
Great Idea, although I don't like text in front of images. Give me some time to think about that (should also be configurable), I'll definitely come back to this PR :)

jneilliii:
Understood, that's why I had added the transparency so the text was more readable.

UnchartedBull:
I had some time and enabled OctoDash to show the preview while printing. See more information here: #570. Please let me know if you're fine with the other implementation, or what you want to be changed in the other PR. Adding this to the config so it defaults to showing the image might be added later.

Issue #235, opened 2019-10-19
Bed Termperature not rendered correctly - possibly when temp is 100C or above


spiff72:
**Describe the bug**
Bed temperature display isn't rendered correctly.

**To Reproduce**
My display resolution is 1024 x 600 (7" screen).
Not 100% sure this is the cause, but I noticed this when attempting my first print with ASA, which has a bed temperature setting of 105 C.
**Expected behavior**
Temps of 100 C or more should render correctly.

**Screenshots**
Image is attached showing the issue.

**General Information:**
 - Hardware: Raspberry Pi 3B+, 1024 x 600 captouch screen
 - OS Info: Raspbian Buster
 - OctoDash Version: Latest Version?  Not sure where to check this

**Additional context**
![20191019_152107](https://user-images.githubusercontent.com/11482743/67150381-e00c0600-f284-11e9-8089-059c67fe215a.jpg)

spiff72:
Octoprint info:

OctoPrint 1.3.11 running on OctoPi 0.13.0

spiff72:
Strangely, after the second layer started (and the temp was increased to 110 C, the temp is rendered correctly.
![20191019_153435](https://user-images.githubusercontent.com/11482743/67150497-21e97c00-f286-11e9-8bf8-7e51811554a3.jpg)

spiff72:
One last comment for now.  I noticed when the bed temp sagged to 109 C the rendering issue came back (sorry - no pic).  So perhaps the rendering is incorrect when temp is between 100 and 109 C (inclusive)?

UnchartedBull:
No it’s just I never thought that the printed bed does never heat up to over 100C, that’s why I only tested til 99, which worked. I‘ll make sure to include the fix.

110 works because it is not as wide as 150. 101 for example will work as well.

spiff72:
Ah - that makes sense.

spiff72:
Is this closed as of 1.3.2, or is the fix coming in v2?  I think that I am running 1.3.2, but not positive (is there a way to check the version?)...and the rendering problem is still present on my 1024 x 600 display.  Layer counts are getting chopped off at the bottom too.

spiff72:
Here is a photo i took today.
![20191027_170506](https://user-images.githubusercontent.com/11482743/67642042-2ba25d80-f8de-11e9-898c-aa86dd9024db.jpg)

UnchartedBull:
[...]

PR #563, opened 2020-04-12
Add sdcard support

### LINKED ISSUE #481: Unable to view SD CARD files plugged into Prusa MK3S
**Describe the bug**
Frist Off i would liek to say what a wonderfull product. The UI i just amazing. However i am am having issue.

I am unable to view files that are on the SD CARD plugged into my prusa mk3s.

Launch the OCTOPRINT browser UI shows the SD CARD FILES listed. Octodash is unable to see these files.

send the command

octopi.local/api/files?recursive=true 
X-Api-Key: *************

Returns a json with all the files listed.

If i generate a folder from OCTOPI this is displayed correctly in OCTODASH

Running the get file command again gives back all files. This time however it shows  the TEST folder as having a origin as "local".

This would suggest that files with origin of "sdcard" are unable to be seen by octodash.

**Expected behavior**
SD files should show in octodash as they do in octoprint

**Screenshots**
image

**General Information:**
Rasberry PI 3B+ with Offical 7 inch Touch screen
Octoprint Version 1.4.0
Octopi version 0.17.0
Octodash version 1.4.0

harleyg321:
Adds more complete support for browsing/printing from the sdcard in octodash.

If there are files available from both local and sdcard, then a fake top level folder is shown to select between local and sdcard. Browsing inside those folders then continues as normal and the path starts with either '/local' or '/sdcard'.

If there are only local or sdcard files, then only the relevant folder is shown. For example if sdcard support is not enabled in octoprint, then you will just see a list of local files as usual. Only difference is the path would now show '/local' at the start instead of '/'. Browsing up to '/' is prevented to keep the user confined to local files.

Also cleans up some of the changes in my previous sdcard commit (#542). Rather than passing through null values to the view, if certain values are not available then they are simply not added to the object. The view then checks if the property exists before displaying certain elements that may not always be available.

Resolves #481


UnchartedBull:
That looks quite impressing. Currently on mobile, I’ll make sure to check this out in detail tomorrow!

src/app/job-status/job-status.component.html
```diff
@@ -5,8 +5,8 @@
         <div class="job-info__progress-percentage">{{ job.progress }}<span style="font-size: 40%">%</span></div>
     </div>
     <span class="job-info__filename">{{ job.filename }}</span> <br />
-    <span class="job-info__filament" *ngIf="job.filamentAmount !== null">{{ job.filamentAmount }}g Filament</span> <br />
-    <span class="job-info__time"><span *ngIf="job.timeLeft.value !== null"><span
+    <span class="job-info__filament" *ngIf="job.hasOwnProperty('filamentAmount')">{{ job.filamentAmount }}g Filament</span> <br />
+    <span class="job-info__time"><span *ngIf="job.hasOwnProperty('timeLeft')"><span
```
UnchartedBull, line 9: that is way cleaner 👍

src/app/job-status/job-status.component.html
```diff
@@ -30,19 +30,24 @@
...
-        <span class="job-info__print-details-finish-state">
-            filament will be used
+        <span *ngIf="job.hasOwnProperty('filamentAmount')">
+            <fa-icon [icon]="['fas', 'dharmachakra']" class="job-info__print-details-icon"></fa-icon>
+            <span class="job-info__print-details-value">
+                {{ job.filamentAmount }}g
+            </span>
+            <span class="job-info__print-details-finish-state">
+                filament will be used
+            </span>
```
UnchartedBull, line 50: this will just be a big empty space if no GCode Analysis is available right?

src/app/files.service.ts
```diff
@@ -100,21 +134,25 @@ export class FilesService {
...
-                .get(this.configService.getURL('files/local' + filePath), this.configService.getHTTPHeaders())
+                .get(this.configService.getURL('files' + filePath), this.configService.getHTTPHeaders())
                 .subscribe(
                     (data: OctoprintFilesAPI): void => {
-                        let filamentLength = 0;
-                        _.forEach(data.gcodeAnalysis.filament, (tool): void => {
-                            filamentLength += tool.length;
-                        });
+                        if (data.gcodeAnalysis) {
+                            var filamentLength = 0;
```
UnchartedBull, line 141:
```suggestion
                        let filamentLength = 0;
                        if (data.gcodeAnalysis) {
```

Is there a need for filamentLength to be function scoped?

src/app/job-status/job-status.component.html
harleyg321: yes, it just hides the estimated completion time and filament weight required since they aren't available, which does leave quite a bit of empty space.

src/app/files.service.ts
harleyg321: good point, fixed both occurrences

spiff72:
Just noticed this when searching issues for "printing with sd card".  I am printing some stuff now from SD card because I now think I get better print quality on fast prints (due to the limitations of the serial interface between pi and printer).

I get a constant flashing error dialog stating "Can't retrieve dialog statys", and another message pops up very briefly while it is doing this.  I recorded it with my phone in slow motion and figured out that it states can't retrieve thumbnail in this short bursts.

Will this fix this issue?

spiff72:
Note: It also messes up formatting of the temps and fan speed. because it says "sd" at the bottom.

harleyg321:
Yes should do, that issue was first fixed in #542 and then this PR refines that a bit further. Both merged after v1.4.1 though so you should get the fixes in the next release.

Though I didn't get any issues with the 'printing from sd' message when testing on the official raspberry pi touchscreen. Could you create a new issue for that please, including which screen you are using and its resolution and/or aspect ratio if known, so we can look into that a bit more.

spiff72:
Sure - here is the info on my screen:
My display resolution is 1024 x 600 (7" screen).
The screen is this one:
https://www.amazon.com/gp/product/B07L6WT77H/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1

spiff72:
I had another issue post with this same screen before (regarding the temperature formatting being "off" when bed temp was 100 or higher).  
https://github.com/UnchartedBull/OctoDash/issues/235

PR #570, opened 2020-04-14
Show preview while printing (click on percentage)

UnchartedBull:
This while show the thumbnail of the current print if you click on the percentage. The choice should be remembered until a restart of OctoDash happens.

Demo:
Screen-Recording-2020-04-14-at-1


jneilliii:
That's totally cool and was what I tried implementing originally, but I couldn't figure out how to get beyond the control page on click.

UnchartedBull:
That was a pain, did a little sneaky workaround which should work for now, but I may revisit that later.

PR #601, opened 2020-04-27
Additional Thumbnail Support

### LINKED ISSUE #566: Display preview images from Prusaslicer gcode
It is nice that OctoDash displays preview images from .ufp files, but I prefer Prusaslicer. Prusaslicer from version 2.2 up generates preview images for the Prusa Mini. But if you change with a text editor the line thumbnails = to thumbnails = 16x16,220x124 in the printer settings file, then you get the same images in the gcode file for every printer. I just tested it. It looks like the images are uuencoded.
Would it be possible to show this images just like the ones from the .ufp files?

jneilliii:
Changes necessary to support both PrusaSlicerThumbnails and UltimakerFormatPackage plugin's extracted thumbnails, resolves #566

### DIFF
--- a/src/app/files.service.ts
+++ b/src/app/files.service.ts
@@ -163,7 +163,7 @@ export class FilesService {
                                 : {}),
                             thumbnail: data.thumbnail
                                 ? this.configService
-                                      .getURL('' + data.thumbnail + '')
+                                      .getURL(data.thumbnail)
                                       .replace('/api/', '/')
                                 : 'assets/object.svg',
                         } as unknown) as File;
--- a/src/app/files.service.ts
+++ b/src/app/files.service.ts
@@ -161,11 +161,11 @@ export class FilesService {
                                       filamentWeight: this.service.convertFilamentLengthToAmount(filamentLength),
                                   }
                                 : {}),
-                            thumbnail: data.path.endsWith('.ufp.gcode')
+                            thumbnail: data.thumbnail
                                 ? this.configService
-                                      .getURL('plugin/UltimakerFormatPackage/thumbnail/')
-                                      .replace('/api/', '/') + data.path.replace('.ufp.gcode', '.png')
-                                : undefined,
+                                      .getURL('' + data.thumbnail + '')
+                                      .replace('/api/', '/')
+                                : 'assets/object.svg',
                         } as unknown) as File;
                         resolve(file);
                     },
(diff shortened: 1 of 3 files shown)


src/app/files.service.ts
```diff
@@ -161,11 +161,11 @@ export class FilesService {
                                       filamentWeight: this.service.convertFilamentLengthToAmount(filamentLength),
                                   }
                                 : {}),
-                            thumbnail: data.path.endsWith('.ufp.gcode')
+                            thumbnail: data.thumbnail
                                 ? this.configService
-                                      .getURL('plugin/UltimakerFormatPackage/thumbnail/')
-                                      .replace('/api/', '/') + data.path.replace('.ufp.gcode', '.png')
-                                : undefined,
+                                      .getURL('' + data.thumbnail + '')
```
UnchartedBull, line 166:
```suggestion
                                      .getURL(data.thumbnail)
```

I think we don't need the concatenation anymore.

src/app/files.service.ts
```diff
@@ -191,11 +191,11 @@ export class FilesService {
                 .get(this.configService.getURL('files' + filePath), this.configService.getHTTPHeaders())
                 .subscribe(
                     (data: OctoprintFilesAPI): void => {
-                        let thumbnail = data.path.endsWith('.ufp.gcode')
+                        let thumbnail = data.thumbnail
                             ? this.configService
-                                  .getURL('plugin/UltimakerFormatPackage/thumbnail/')
-                                  .replace('/api/', '/') + data.path.replace('.ufp.gcode', '.png')
-                            : undefined;
+                                  .getURL('' + data.thumbnail + '')
```
UnchartedBull, line 196:
```suggestion
                                  .getURL(data.thumbnail)
```
same as above.

src/app/files/files.component.html
```diff
@@ -70,7 +70,7 @@
         <span class="file__name">{{ fileDetail.name }}</span>
         <span class="file__directory">{{ fileDetail.path }}</span>
         <span class="file__creation-date" *ngIf="fileDetail.hasOwnProperty('date')">{{ fileDetail.date }}</span>
-        <img src="{{fileDetail.thumbnail ? fileDetail.thumbnail : 'assets/object.svg'}}" class="file__render"/>
+        <img src="{{fileDetail.thumbnail}}" class="file__render"/>
```
UnchartedBull, line 73:
```suggestion
        <img [src]="fileDetail.thumbnail" class="file__render"/>
```

If we don't have a ternary operator in there anymore, we can use [src] here I think.

src/app/octoprint-api/filesAPI.ts
```diff
@@ -14,6 +14,7 @@ export interface OctoprintFilesAPI {
     statistics?: object;
     type: string;
     typePath: [string];
+    thumbnail?: [string];
```
UnchartedBull, line 17: Is this really an array?

src/app/files.service.ts
jneilliii: I initially tried this but was getting errors, but I think that's because of `[string]`, will update and verify.

src/app/octoprint-api/filesAPI.ts
jneilliii: No, I just wasn't paying attention...which is why my other calls were causing errors hence the concatentation...

jneilliii:
Requested changes committed and verified working, thanks for the pointer on that string/array thing.

PR #643, opened 2020-05-15
Remember Feed- and Flow-Rate

### LINKED ISSUE #640: Current feedrate not correctly shown in the adjust menu
While printing when using the adjustment the current feedrate is not correctly shown in adjustment menu. I'm not sure if this is an OctoDash bug or a lack of capability in OctoPrint.

**To reproduce:**
1. Start print.
2. Touch screen and click on adjust. 
3. Change feedrate to 75% and change bed temperature from 60 to 65C
4. Correct expected behaviour observed on printer, printer slows down and temp increases. 
5. Touch screen and click on adjust. 

- Expected behaviour: Feedrate shows 75% and Bed temperature shows 65C
- Actual behaviour: Feedrate shows 100%, and Bed temperature correctly shows 65C.

6. Open Octoprint and click on control
- Expected behaviour: Feedrate in Octoprint's control menu shows 75%
- Actual behaviour: Feedrate in Octoprint's control menu unchanged at 100%. 

7. Change feedrate in Octoprint's control menu to 80%
8. Correct expected behaviour observed on printer, printer speeds to 80%.
9. Touch screen and click on adjust.

- Expected behaviour: Feedrate shows 80%
- Actual behaviour: Feedrate remains unchanged shows 100%

At present there doesn't seem to be any coupling between the feedrate in OctoDash and the feedrate in Octoprint, however both correctly send the M203 command to the printer. 

If Octoprint does not provide access to the current feedrate value in their API it may be worth at least tracking what it was last set to on the Octodash display. 

**EDIT: Additional implication of this bug:**
Not possible to set feedrate back to 100% as the feedrate always appears to be 100% when going into the adjustment menu the 100% is not sent to the printer when hitting save. I was only able to get it to 99% or 101%

UnchartedBull:
Fixes #640

### DIFF
--- a/src/app/print-control/print-control.component.ts
+++ b/src/app/print-control/print-control.component.ts
@@ -19,7 +19,12 @@ export class PrintControlComponent {
     public feedrate: number;
     public flowrate: number;
 
-    public constructor(private jobService: JobService, private printerService: PrinterService) {}
+    public constructor(private jobService: JobService, private printerService: PrinterService) {
+        this.temperatureHotend = 0;
+        this.temperatureHeatbed = 0;
+        this.flowrate = 100;
+        this.feedrate = 100;
+    }
 
     public isClickOnPreview(event: MouseEvent): boolean {
         const previewSwitchMin = window.innerWidth * 0.08;
@@ -99,10 +104,6 @@ export class PrintControlComponent {
     }
 
     private loadData(): void {
-        this.temperatureHotend = 0;
-        this.temperatureHeatbed = 0;
-        this.flowrate = 100;
-        this.feedrate = 100;
         this.printerService
             .getObservable()
             .pipe(take(1))


thegarbz:
Erm sorry to raise this point again even though you just corrected it. On the discussion with the OctoPrint developer he was completely resistant to maintaining a value that is ephemeral in Octoprint. Meaning that if the firmware can't confirm what the rate currently is the software shouldn't attempt to memorise it. His view has actually been to blank out the feedrate and flowrate when the user leaves the tab or after the user submits a new value to the printer. 

For consistency sake this may be a better solution for Octodash as well. Frankly I'm okay with either option. But Octoprint's solution does reflect that Octoprint doesn't actually know what the printer is doing. 

But anyway this is just for consideration.

PR #662, opened 2020-05-19
Change thumbnail to [hidden]?

noxhirsch:
Hi,
this is a rather unusual PR: I want to make some pretty big design changes, but while I of course can publish the custom CSS file, I don't think you don't want to implement it as the main style ;)
One of my big changes is to show the thumbnail all the time (and bigger). Sadly with Angulars "*ngIf" the thumbnail is not hidden but completely gone. So I made these changes which look the same for most of the users, but give the rest the ability to overwrite the `display:none;` and always show the thumbnail.
I hope you consider merging it - but if not, I can understand that. Or maybe you have a better/cleaner solution?

By the way: Here's the first change I made. The file popup now is fullscreen (to better use the small space of my 4" display) and has a translucent blurred background and a big thumbnail.
screenshot

### DIFF
--- a/src/app/job-status/job-status.component.html
+++ b/src/app/job-status/job-status.component.html
@@ -2,9 +2,9 @@
     <div class="job-info__progress-ring" (click)="showPreview = !showPreview">
         <round-progress [current]="job.progress" [max]="100" [stroke]="25" [rounded]="true" [responsive]="true"
             [color]="'#44bd32'"></round-progress>
-        <div class="job-info__progress-percentage" *ngIf="!job.thumbnail || !showPreview()">{{ job.progress }}<span
+        <div class="job-info__progress-percentage" [hidden]="job.thumbnail && showPreview()">{{ job.progress }}<span
                 style="font-size: 40%">%</span></div>
-        <img *ngIf="job.thumbnail && showPreview()" class="job-info__progress-preview" [src]="job.thumbnail" />
+        <img [hidden]="!job.thumbnail || !showPreview()" class="job-info__progress-preview" [src]="job.thumbnail" />
     </div>
     <span class="job-info__filename">{{ job.filename }}</span> <br />
     <span class="job-info__filament" *ngIf="job.hasOwnProperty('filamentAmount')">{{ job.filamentAmount }}g Filament</span> <br />
--- a/src/app/job-status/job-status.component.scss
+++ b/src/app/job-status/job-status.component.scss
@@ -33,6 +33,11 @@
             display: block;
         }
     }
+    
+    & [hidden] {
+        display: none;
+    }
+
 
     &__filename {
         font-size: 6vw;


UnchartedBull:
I was actually thinking about making the custom-styles publishable and create some kind of a "theme-store" for OctoDash. Since we'll have the custom styles available with v2, I might as well make them easily shareable. Might be something for v3 :) I might come back to you for that, if that is fine with you.

Regarding your change: You want to accomplish, that the preview image in the job-status is always shown and that you don't have to click on the percentage, correct? The problem with hidden here is, that Angular will still try to load the image, since it is just hidden, which will result in errors being thrown, that could potentially lead to a memory leak. My idea here would be to introduce a new config parameter to switch the normal behaviour around. So on default it will show the preview image and you will have to click on it to show the percentage. Would this also work for you? If you like you can do that as well, otherwise I could take over that task. Just let me know :)

noxhirsch:
> I was actually thinking about making the custom-styles publishable and create some kind of a "theme-store" for OctoDash. Since we'll have the custom styles available with v2, I might as well make them easily shareable. Might be something for v3 :) I might come back to you for that, if that is fine with you.

That sounds like a great idea! 👍

> Regarding your change: You want to accomplish, that the preview image in the job-status is always shown and that you don't have to click on the percentage, correct? 

Partly: I'd like to show both the progress and the thumbnail. Maybe the solution could be to use `[hidden]` to hide the thumbnail when existing and additionally use `*ngIf` to hide it when no thumbnail available?
Or use the config parameter to show the thumbnail first, as you suggested, but only hide the progress with [hidden] to make it overwritable?
A config parameter to show both would need some bigger changes on the main design, I think (except when showing the progress percentage only very small).
For any config parameters changes it would be great if you could take over the task. I'm not good at TypeScript/JavaScript.

noxhirsch:
Additional idea: For multiple themes it might be useful to have the ability to replace the icons, but I don't know what's the cleanest way to accomplish that.

noxhirsch:
@UnchartedBull Is there a way to tell, on which page e.g. the printer status is shown? Like `.main-menu .printer-status` (which obviously doesn't work)
Than I could design it more individually depending on the available space per page.

UnchartedBull:
Using [hidden] for the progress might be the best idea here then I think. I'll have a look at this soonish. :)

Yes you have surrounding HTML elements for each screen. The exact names are listed in their respective components, but they should look like app-main-menu and app-printer-status. Please note that these are not classes, but rather normal html elements, so the selector needs to look like this: `app-main-menu`.

noxhirsch:
It didn't work directly, because app-printer-status is not inside app-main-menu (like I thought it would be), but I just found out that this works:

```
app-main-menu + app-printer-status .printer-status{
    background: red!important;
}
app-job-status + app-printer-status .printer-status{
    background: green!important;
}
```

UnchartedBull:
Hi, could you please check, whether https://github.com/UnchartedBull/OctoDash/pull/690 this works for you? You can enable the config parameter in the config in the OctoDash sub directory. I haven't tested it yet, since I don't wanted to start a print. Will test this tomorrow, but it should work just fine.

noxhirsch:
Hi, I'll check that as soon as I'm back from work in about 9 hours 👍

UnchartedBull:
Ok, I quickly checked it now, while a print is running and everything seems to work now. Just needed a little tweaking. Closing this PR now, since it will be replaced by #690. Just let me know over there, whether this works for you. Thanks for the contribution!

PR #690, opened 2020-05-27
add new conig parameter to show thumbnail by default while printing

UnchartedBull:


noxhirsch:
Everything seems to work for me! 👍 Thank you for your help!

PR #856, opened 2020-07-29
Fix issue #849; Respect the OctoPrint axis invert flags

### LINKED ISSUE #849: Axis controls don't match OctoPrint (ie: inverted setting)
When using the control screen to move the toolhead X/Y/Z the controls don't respect the axis inversion settings in OctoPrint

1. Go to OctoPrint sttings and choose to invert the movement of an axis.
2. In OctoPrint use the controls to move the tool head and note that the controls respect the inverted flag of any axis.
3. In OctoDash use the controls to more the tool head and note that the controls do not respect the inverted flag.

 - Raspberry Pi 3B+
 - OctoPi v0.17.0 (Raspbian GNU/Linux 10)
 - OctoDash v1.5.0
 - OctoPrint v1.4.0

pfemiani:
I went ahead and took a stab at fixing the issue where the OctoPrint axis inversion setting was not being respected in the in the printer control view.

- Created PrinterProfileService
- Implemented getDefaultPrinterProfile()
- When Control view loads it gets the default printer profile, and subsequent axis moves take the "inverted" setting into account

Fixes #849.

### DIFF
--- a/src/app/printerprofile.service.ts
+++ b/src/app/printerprofile.service.ts
@@ -1,8 +1,7 @@
 import { HttpClient, HttpErrorResponse } from "@angular/common/http";
 import { Injectable } from "@angular/core";
 import { Router } from "@angular/router";
-import { Observable, Observer, Subscription, timer } from "rxjs";
-import { shareReplay } from "rxjs/operators";
+import { Subscription } from "rxjs";
 
 import { ConfigService } from "./config/config.service";
 import { NotificationService } from "./notification/notification.service";
--- a/src/app/printerprofile.service.ts
+++ b/src/app/printerprofile.service.ts
@@ -31,8 +31,7 @@ export class PrinterProfileService {
       this.httpGETRequest = this.http
         .get(this.configService.getURL("printerprofiles/_default"), this.configService.getHTTPHeaders())
         .subscribe(
-          (data: OctoprintPrinterProfileAPI): void => {
-            const printerProfile: OctoprintPrinterProfileAPI = data;
+          (printerProfile: OctoprintPrinterProfileAPI): void => {
             resolve(printerProfile);
           },
           (error: HttpErrorResponse): void => {
--- a/src/app/printerprofile.service.ts
+++ b/src/app/printerprofile.service.ts
@@ -14,27 +14,15 @@ import { OctoprintPrinterProfileAPI } from "./octoprint-api/printerProfileAPI";
   providedIn: "root",
 })
 export class PrinterProfileService {
-  private printerStatusService: PrinterService;
   private httpGETRequest: Subscription;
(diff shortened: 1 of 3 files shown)


src/app/printerprofile.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    return new Promise((resolve): void => {
+      if (this.httpGETRequest) {
+        this.httpGETRequest.unsubscribe();
+      }
+      this.httpGETRequest = this.http
+        .get(this.configService.getURL("printerprofiles/_default"), this.configService.getHTTPHeaders())
+        .subscribe(
+          (data: OctoprintPrinterProfileAPI): void => {
+            const printerProfile: OctoprintPrinterProfileAPI = data;
+            resolve(printerProfile);
```
UnchartedBull, line 36:
```suggestion
          (printerProfile: OctoprintPrinterProfileAPI): void => {
            resolve(printerProfile);
```

The observable already maps to the correct type, so we can save that variable as far as I can tell.

src/app/printerprofile.service.ts
```diff
@@ -0,0 +1,69 @@
...
+          (error: HttpErrorResponse): void => {
+            const printerProfile: OctoprintPrinterProfileAPI = {
+              name: "_default",
+              model: "unknown",
+              axes: {
+                x: { inverted: false },
+                y: { inverted: false },
+                z: { inverted: false }
+              }
+            };
```
UnchartedBull, line 47: I would probably remove this and instead reject the observable (just catch it in control.component and do nothing there). Since you already added the default profile in the control.component returning nothing here might be better. This will either keep the default profile or, if a prior request was successful, keep the last retrieved printer profile, which should be better than the default. Let me know if you want to work on that, or whether I should fix that :).

UnchartedBull:
Thank you. Will be released with the next version :)

pfemiani:
@UnchartedBull ... I have no idea how I missed this while testing, probably because I was mostly looking at inputs/outputs, and visually testing primarily with Z axis... but the Y axis controls in OctoDash seem to actually be sending the wrong direction. If you click the up arrow, it is configured for a "-" direction, and down arrow configured for "+". However, in OctoPrint if you click the up arrow it moves Y in positive direction (assuming axis inversion is not enabled). I verified debugged in and verified X and Z axis both move in correct direction, and all 3 axis correctly respect the inversion setting, just the Y axis controls are backward. Let me know if you would like me to create an issue or another PR to fix this, or if its easier for you to just knock it out since it should simply be switching the "-" and "+" for the Y axis in control.component.html.

Thanks in advance! I'm really loving OctoDash!

UnchartedBull:
I'll have a look at it once I'm back :)

Issue #921, opened 2020-08-19
OctoPrint plugin for OctoDash config


TheSin-:
Just wanted a separate entry for creating an octoprint plugin for config handling on octodash to avoid using a keyboard.

More info to come on this but it could also be using to gather info that might not be in the API, like and API extension in the future.

UnchartedBull:
Rudimentary support is already done #856, so the printer profile is known to OctoDash, it currently just ignores everything except for axis inversion. It would make sense to move this to a separate service though, since it would be used by multiple components.

UnchartedBull:
An actual plugin would be super useful. I don't know whether this is possible (I mean it is definitely possible, but whether it is allowed as a plugin) is to bootstrap the OctoPrint Installation to the plugin installation, that would also make the setup a lot easier :)

UnchartedBull:
Ok maybe another thing from the top of my head. One of the biggest changes planned for v3 is the switch to OctoPrints socket interface (https://docs.octoprint.org/en/master/api/push.html), which should marginally reduce load, but finally would make DLP optional as well (since Z-Height is available there). The problem is that you need to create a session once (the user needs to login for that). What I just thought about is something like a handshake process - The user would start OctoDash and would click an authenticate button or something like that, this would then result in a popup from the plugin which asks the currently logged in user to grant permission to OctoDash thus creating a token for OctoDash. Manual input should still be available though, but if that works out there is no real need for installation wizard anymore (or only a really small one with only checkboxes and maybe number inputs), since most of the other data can be retrieved from the Printer Profile. That would be super nice. Let me know if you want to investigate this further, or whether I should check if that's possible.

TheSin-:
I do remember ready you comment on an other ticket about switching to push.  Are you far along on that?  Should make an octoprint api service that other services can access this would keep all eh calls to the api together and clean?  Making it easier to switch the calls later to push since we would just use function calls internally?  My current Angular project that is what I'm doing since the API isn't complete and I figure it'll be easier to change since I can control the input and output in the function no matter what changes.

TheSin-:
hmm that's a good plan honestly.  I was kinda thinking the same, if octodash started up with an ip only on the display.  Then in octoprint plugin you enter the ip to initiate the setup, that could exchange the initial data we require.  the next thing on the script could be a 6 digit key just like bluetooth that would just confirm the right devices on the same network.  Incase there is more then one, between the ip and the key I'm sure we could trust it for the setup.

[...]

Issue #762, opened 2020-06-24
Use onscreen keyboard or other input method to define config file


MatthewCroughan:
It would be great if the config.json file were definable entirely by touching the screen, or a web interface so that noobs could easily configure this forever, without ever having to interface with the CLI.

MatthewCroughan:
https://www.npmjs.com/package/electron-virtual-keyboard

UnchartedBull:
Good idea. I like the design of this one a little better though: https://www.npmjs.com/package/simple-keyboard. Leaving both here for now.

dintxaustip:
Hola a todos. Soy un absoluto novato, pero después de 20 intentos (y no exagero) he conseguido instalar octodash en mi pantalla MHS-3.5. Lo primero que conseguí fue ver la interfaz de raspbian en mi lcd. Cambié en Raspi-config para que se cargara automáticamente al arrancar y actualmente me muestra el saludo de Octodash para configurarlo. Problema: no tengo teclado en pantalla para completar la configuración. Entiendo que se puede instalar un teclado virtual, pero no sé cómo hacerlo. Me gustaría que apareciera en pantalla cuando pretenda rellenar los huecos de texto. Utilizo Putty para meter toda la información. Qué órdenes debería copiar?. Agradezco de antemano su atención y espero me puedan ayudar. Gracias

UnchartedBull:
I can't exactly can help you with the on-screen keyboard, but you can also use xdotools as outlined here: https://github.com/UnchartedBull/OctoDash/wiki/Setup-&-Settings#setup-without-keyboard. There are also multiple tutorials available on how to use xdotools around the internet :)

Also temporarily attaching a keyboard (only needed for setup) is an option as well :)

TheSin-:
Just as an idea what is octodash had an octoprint plugin, so you could enter everything (except the api info) via the plugin (web ui), you could also use the plugin to add any info you want form octoprint, printer name, extruder count and setup, etc etc

Just an idea, I was working on adding the virtual KB but I don' think it's going to work since the display is so small it's a difficult task, this might make the best of both worlds.  As for the API info you could read it from a special file on /boot so when you setup octopi and you need to mode the spa info text file we could have an octodash txt file there as well that get copied into place on boot.

Again these are just ideas to try to make this great project easy to setup.

UnchartedBull:
[...]

PR #954, opened 2020-08-29
Reduce input during on-boarding

TheSin-:
- Add mdns and ngx-electron plugin
    - Add electron-rebuild to dev depends so that electron can be rebuilt with the installed node
    - Fix setTimeout and setInstance to with with Node modules (required for mdns)
    - Reduce the version on node as electron needs to be on the same version for this to work
    - Change order in on boarding wizard to select API first
    - Use mdns to detect and list octoprint instances (WIP need to figure out min version required)
    - Auto set Printer name if a detected install if found.

Step 1 in #921


src/app/config/no-config/no-config.component.ts
```diff
@@ -10,27 +12,40 @@ import { Config, ConfigService } from '../config.service';
...
+    'other': {
+      'display': 'Other (> 1.4.0)',
+      'name': 'other',
+      'version': '1.4.0',
+      'url': 'other',
+      'disable': false
+    }
+  };
+  public opInstance: any = this.octoprintNodes['other'];
+  public opApiMsg: string = '';
```
UnchartedBull, line 35: I guess we should use octoprintInstance and octoprintApiMessage here, since the full name is used in other vars as well.

src/app/config/no-config/no-config.component.ts
```diff
@@ -10,27 +12,40 @@ import { Config, ConfigService } from '../config.service';
   styleUrls: ['./no-config.component.scss'],
 })
 export class NoConfigComponent implements OnInit {
-  public page = 0;
-  public totalPages = 6;
+  public page: number = 0;
+  public totalPages: number = 7;
```
UnchartedBull, line 16: the `:number` is redundant here

src/app/config/no-config/no-config.component.ts
```diff
@@ -10,27 +12,40 @@ import { Config, ConfigService } from '../config.service';
...
+  public constructor(private configService: ConfigService, private http: HttpClient, private router: Router, private _electronService: ElectronService) {
     this.configUpdate = this.configService.isUpdate();
     console.log(this.configUpdate);
     if (this.configUpdate) {
       this.config = configService.getCurrentConfig();
     } else {
       this.config = {
         octoprint: {
-          url: 'http://localhost:5000/api/',
+          url: 'http://localhost:80/api/',
```
UnchartedBull, line 48: We should stick with port 5000 here, since that is the official port from OctoPrint. (80 is just added from OctoPi)

src/app/config/no-config/no-config.component.ts
```diff
@@ -137,9 +152,28 @@ export class NoConfigComponent implements OnInit {
...
+        // Compare version to make sure it meets the requirement
+        'disable': false
+      };
+
+      this.octoprintNodes[service.host.replace(/\.$/, '').replace('.', '_')] = node;
+    });
+    browser.on('serviceDown', service => {
+      delete this.octoprintNodes[service.host.replace(/\.$/, '').replace('.', '_')];
+    });
+    browser.start();
```
UnchartedBull, line 173: Can we put this in `main.js`, this would remove the ngx-electron dependency and the icp links are already there, plus we would get around having to add node to the main angular app.

src/app/config/no-config/no-config.component.ts
```diff
@@ -166,12 +200,12 @@ export class NoConfigComponent implements OnInit {
     return true;
   }
 
-  public validateConfig(): void {
+  public async validateConfig(): Promise<any> {
```
UnchartedBull, line 203: Should return `Promise<void>`
```suggestion
  public async validateConfig(): Promise<void> {
```

src/app/config/no-config/no-config.component.ts
```diff
@@ -184,16 +218,41 @@ export class NoConfigComponent implements OnInit {
     this.router.navigate(['/main-screen']);
   }
 
-  public increasePage(): void {
+  public async increasePage(): Promise<any> {
```
UnchartedBull, line 221:
```suggestion
  public async increasePage(): Promise<void> {
```

src/app/config/no-config/no-config.component.ts
```diff
@@ -137,9 +152,28 @@ export class NoConfigComponent implements OnInit {
...
+      this.octoprintNodes[service.host.replace(/\.$/, '').replace('.', '_')] = node;
+    });
+    browser.on('serviceDown', service => {
+      delete this.octoprintNodes[service.host.replace(/\.$/, '').replace('.', '_')];
+    });
+    browser.start();
   }
 
-  public testOctoprintAPI(): boolean {
+  public async testOctoprintAPI(): Promise<any> {
```
UnchartedBull, line 176:
```suggestion
  public async testOctoprintAPI(): Promise<boolean> {
```

src/app/config/no-config/no-config.component.ts
```diff
@@ -184,16 +218,41 @@ export class NoConfigComponent implements OnInit {
...
+    }
     if (this.page === this.totalPages) {
       this.createConfig();
     }
     this.changeProgress();
   }
 
   public decreasePage(): void {
-    if (this.page === 5) {
+    if (this.page == this.totalPages - 1) {
```
UnchartedBull, line 255: 👍

src/app/config/no-config/no-config.component.html
```diff
@@ -41,7 +41,7 @@
       <fa-icon [icon]="['fas', 'smile']"></fa-icon>
     </span>
   </div>
-  <div *ngIf="page === 1">
+  <div *ngIf="page === 3">
```
UnchartedBull, line 44: I think these should be sorted again once everything is complete. Fine for now though, since the whole setup screen needs a revamp once all the changes are in.

src/app/config/no-config/no-config.component.ts
```diff
@@ -10,27 +12,40 @@ import { Config, ConfigService } from '../config.service';
...
   public config: Config;
   public configErrors: string[];
   public configValid: boolean;
   public configSaved: string;
 
-  public octoprintConnection: boolean;
+  public objectvalues = Object.values;
+  public octoprintNodes: any = {
+    'other': {
+      'display': 'Other (> 1.4.0)',
```
UnchartedBull, line 27: I think we can be more inclusive here. I would propose >1.3.5.

src/app/config/no-config/no-config.component.ts
```diff
@@ -184,16 +218,41 @@ export class NoConfigComponent implements OnInit {
...
+            this.config.printer.name = this.opInstance.name;
+          } else {
+            this.config.printer.name = '';
+          }
+        }, err => {
+          this.opApiMsg = 'API Error: ' + this.octoprintConnectionError;
+          this.page = 2;
+        });
+      }
+    }
```
UnchartedBull, line 247: This will query the API 4 times (for each page after 2). I would prefer to just have the check for connection at the very end, where it is currently located. Or are there any benefits to having it here?

UnchartedBull (changes requested):
Looks really good. The only bigger point to discuss here, is whether the mdns part should be moved to the main.js, where we already have full access to the node api, sharing data can be easily done via the electron IPC. And the API Connection check can be at the end I guess. Everything else is just some minor codestyle stuff (I guess the pipeline will also complain about some of that).

src/app/config/no-config/no-config.component.ts
UnchartedBull: Let me know if you want me to do this.

src/app/config/no-config/no-config.component.ts
TheSin-: Sure, I'm lazy I guess ;). I'll update this tonight.

src/app/config/no-config/no-config.component.ts
TheSin-: I like it for consistency, but I can't remove it if it bugs ya.

src/app/config/no-config/no-config.component.ts
TheSin-: oh I didn't know that, maybe we could put a note, but since you official release used is octopi it's make more sense if it worked out of the box I know I got stuck on this a lot in the start.

src/app/config/no-config/no-config.component.ts
TheSin-: Yeah that is beyond me, this took me long enough to figure out ;)

src/app/config/no-config/no-config.component.ts
TheSin-: Didn't know I could do this void there, I know there are other types but I didn't know void was one, this will be handy thanks.

src/app/config/no-config/no-config.component.ts
TheSin-: Did I miss this one?

src/app/config/no-config/no-config.component.html
TheSin-: I need to go retest these I though I tested going all the way through and back 👍 

But that is why I didn't sort it I assume most of this will be gone in step 2.

src/app/config/no-config/no-config.component.ts
TheSin-: sure I had no idea as I haven't finished going through everything, since we plan to use push when was that added?

src/app/config/no-config/no-config.component.ts
TheSin-: it won't cause of line 234.  Once it connects once it'll stop.

src/app/config/no-config/no-config.component.ts
UnchartedBull: Screenshot 2020-08-31 at 15 21 43

Actually bugs VSCode :D

src/app/config/no-config/no-config.component.ts
UnchartedBull: No that was just an: I like that reaction, sorry for the confusion :)

src/app/config/no-config/no-config.component.html
UnchartedBull: Leave that for now, We can go through this once the rest is done. There will be a couple of changes here I guess.

src/app/config/no-config/no-config.component.ts
TheSin-: ahh I use VIM so I don't have that issue, seems like a VSCode issue, but we can remove it since it doesn't hurt anything but my eyes ;)

src/app/config/no-config/no-config.component.ts
UnchartedBull: Oh I missed that if statement. Sorry.

src/app/config/no-config/no-config.component.ts
UnchartedBull: Ok I'll take that one :)

src/app/config/no-config/no-config.component.ts
UnchartedBull: A note would be a good idea yeah. Although localhost:5000 should also work on OctoPi. OctoPi is just creating a reverse proxy from :80 to :5000 AFAIK.

TheSin-: Odd it doesn't work on my Octodash install or my 2 other octopi installs, maybe I'm doing something wrong, I'll look into it.

Either way I'll add a note for it.

src/app/config/no-config/no-config.component.ts
TheSin-: looks like GitHub botched up my comments.  This comment was for an other one.

For this one I mentioned something about when push was added to the api since we will be using that.  I still need to add a version compare lib like semver or something.  I can do that tonight too.

TheSin-:
Because of my changes to post install I might need to move the electron-rebuild dev depend into depends?  I'm really new to electron.

TheSin-:
Done addressing things, I'm going to see if I can figure out why CI is falling next.

UnchartedBull:
Sorry for the spam commits, just wanted to test whether I can push to the PR :)

TheSin-:
no problem I've been learning what can be done too ;)

TheSin-:
The new Travis fails is due to the changes in .eslintrc.js

Which is out side of my knowledge

https://travis-ci.com/github/UnchartedBull/OctoDash/jobs/379614533

UnchartedBull:
Yeah I'll fix those :)

TheSin-:
I think we committed at the same time and some how mine because forced even though I didn't force it :\

UnchartedBull:
No problem, you only did linter stuff right? I'll probably just force push over that, if that's ok :)

TheSin-:
Sorry was driving home, nice changes I'm going to test it after dinner.

TheSin-:
it doesn't test the connect, thus I can't pre populate the rest of the config anymore nor will it warn.  The changes look much better but they user experience isn't as good.

TheSin-:
It's also very slow for initial detection now, and looking at it I don't think it's continue to search so if you turn it one once you realize you need it it'll be too late.

UnchartedBull:
It currently doesn't test the connect no. The idea I'm having currently is to now trigger the API Connection with the popup thingy. This will be easiest if we use the OctoPrint JS Client, which would also allow for super easy socket connection down the line. It's just a big pain to include a dynamically loaded JS library into a TS Angular application. Once that it working we would have the OctoPrint instance + an authorised client without having to put any text input anywhere (of course there will be fallbacks).

The idea for now is:
- Page 1: List nodes / enter node manually
- Page 2: Authorize via popup in WebUI / enter token manually
- Page 3 and onwards: a fully authenticated OctoPrint Client will be ready and can be used to pre-populate most of the values.

I don't know if that makes sense, but I hope so. It's 1AM here and I've been coding since 9AM (including work though), so my brain is a bit mushy right now :D But would that work for you? I know I shifted around a lot of code, but this also means that we have node.js back out of the main Angular app.

Initial detection is just slow due to Angulars Change Detection, we could trigger those manually and have the data almost instantly. I might add that tomorrow. It will continue to search. Each time you leave page 1 the browser will be stopped to not spam the whole network with mdns requests, but if page 1 get's opened again it will search again. I tried that with turning the WiFi off, which makes the instance disappear, but it comes back once the connection is reestablished.

UnchartedBull:
Ok I updated the change detection, should be as fast as before now. Sorry again if I crushed your plans here, but I think we do it now, we should do it right and since we need that library later anyways we might include it now. If you want some changes let me know and if you have more questions let me know as well :) Otherwise enjoy your dinner!

TheSin-:
ahh that was going to be more phase 2 ;). You are doing it in one sweep, well I'm excited to see it, your UI design is beautiful I love that part a lot.

TheSin-:
That's totally okay, I just didn't know you were going to crush it all out at once.  I didn't want to make the PR too large.  I've done that in other projects and it's been over 10 years and it's still not accepted.  So I'm shy about it now, but that is the end goal so I'm more then happy to help get us there any way I can cause I think that will be the ultimate user experience!

UnchartedBull:
I didn't plan to do it from the beginning but then I kinda got lost in it. I'll probably can do that tomorrow evening/night :)

I'm more than ok with large PRs for OctoDash (I do them myself quite frequently), so don't be shy there :) it might just take a little longer for me to review them.

And yes I'm already loving that discovery so much. Makes it a lot easier for not so tech savvy users and is also a great learning for me.

TheSin-:
Well I'm excited to say the least, I'm going to try to tackle the push conversion next, but I'll wait for this PR to be done so I don't need to make a branch ;). Let me know if you need anything else form me on this other then testing.

TheSin-:
looking at the changes we might not need any of the electron-rebuild stuff, but I haven't tested that yet, I'll merge in master if you'd like so that there aren't any conflicts.

UnchartedBull:
Sure! I hope to get this done by tomorrow, after that you should have the authorized client available and then you can pre-fill the values based on printer profile / other information. Rebuild probably isn't needed anymore. I can remove that :)

TheSin-:
I'll test it tonight and remove it that way you can focus on more useful parts that I can't do ;)

TheSin-:
No problems removing it all, just going to see if the travis changes were needed for mdns in general or just the rebuild.

TheSin-:
Just to keep track in my testing one other piece the tis missing from my code is that you do not reset values on back.

so if I pick an Instance from the list, then go back cause I changed my mind and hit enter manually, it's the info from the last Instance I tried instead of the default local value which it should be IMHO.

UnchartedBull:
Ok quick update: the reset if manually entered you're talking about has been implemented. Checking that OctoPrint is running on the specified host:port is also done and the library is loaded correctly. So there is just the authentication workflow left. I hoped that I could do that today, the freakin' jQuery issue took me a good hour to solve though. Gotta postpone the authentication to tomorrow, sorry. After that everything should be done from my side for now I guess. Maybe some adjustment and deletion of non-necessary fields (due to printer profiles.)

My game plan here is to merge this into a separate branch for now and then do one sweep from that branch into master once push updates have been implemented and the setup has been finalised. Will be released with 2.1.0 then I guess, so there'll probably be one or two patch release before this makes it into the final release. Does that work for you?

TheSin-:
Looking good...

There was a small bug that bugged me, on page 1 it defaults to search but if you clicked next it used localhost even though it wasn't listed or selected.  I removed the next button forcing the user to select an Instance or select local to correct the flow and avoid errors.

TheSin-:
There was also a small bug that allowed you to skip past page 0 which I corrected.  Now that I think about it I should have done it in the functions as well to be safe, an other commit inc

TheSin-:
nice that is how I was going to move it but I didn't want to move the click incase there was a reason it was on the div and not the span.

UnchartedBull:
Ok, great news. Seems like everything is working now. Connection is being established and successfully authorized (for now it just retrieves the printer name). No keyboard needed anymore for the setup 🎉 . Let me know if you find any bugs / have some ideas for further changes, this should be ready to merge otherwise :)

Screenshot 2020-09-04 at 18 33 09

TheSin-:
Awesome can't wait to try/test it. I'm just heading out for a bit, I'll test it when I return though!

TheSin-:
I'm searching like made but in electron is there no way to show the number input spinner ?  That would be the best to keep it 100% touch only.

TheSin-:
Maybe something like http://angular-step-input.10kb.nl ?

TheSin-:
I have an idea for the numeric inputs that I"m going to try to work on this week.

TheSin-:
This is my idea, but the two way data bind is really really slow to update and I can't figure out why.  It needs to be much faster to be usable, is this due to electron?

TheSin-:
It's something with the IPC calls, if I set page to 4 and start it (skip the api calls) then it's super responsive and very nice to use.  If I start at page 0 then it's pretty much unusable.  Since I really do not understand the IPC stuff I think you might need to debug it.

UnchartedBull:
Hi, sorry last week has been kinda busy. I'll have a look at that today or tomorrow :) Is there an option to enter text manually though? I think it will be really hard to get to the right number on 3.5" device. I guess we can change that now as well but all other text fields should be updates as well to the new input method. That will be tracked in #762 though.

TheSin-:
I figured we could make the number display (span) into an input, or we could have a toggle to go between the 2.  But I didn't want to work on it if I couldn't get it working at all.

UnchartedBull:
Just tried it and it works just fine for me. No noticable lag at all (I started on page 0). It looks nice and certainly is a super clean way of inputting a number. The main problem here is, that some (possibly many) people are using 3.5" GPIO displays with something close to 1 or 2Hz refresh rate. This makes nice gestures for essential stuff almost impossible.

I redid those so we have + und - buttons (hold them to go up/down faster). I'm fairly happy with that, but please provide feedback on that. Sorry for crashing everything again. This PR is ready to be merged into master from my side now (if the pipeline passes). Let me know what you're thinking :)

TheSin-:
I'll try them in a few minutes, just have a few meetings first.  But if it works well I think that is the last of it, I'm excited!!  This is a great update indeed!

UnchartedBull:
Sure take your time, doesn't really matter if this PR stays open for another day or two :)

TheSin-:
- Looks good, is the alignment issue on purpose or is it just cause I'm running it on my Mac?

- Are we going to add it to the Interval on Page 5 as well?  Just to keep things consistent?

Screen Shot 2020-09-15 at 8 51 28 AM

UnchartedBull:
That layout issue shouldn't happen. Some weird CSS thing going on with the input I guess. Should be fixed now though. The value interval won't receive the buttons since it will be removed soon (once the push updates are in place there is really no need for that anymore), so we can safe ourself the hassle of adding the controller there to.

TheSin-:
yeah I was on the fence about cause really it shouldn't normally be changed.  I'll try the new changes but otherwise I'm gtg on this, it's amazing work!

TheSin-:
Odd it fixed the large version but not the smaller version

Screen Shot 2020-09-15 at 9 10 55 AM

Screen Shot 2020-09-15 at 9 11 04 AM

UnchartedBull:
Damn, though I had fixed it. Will try again later today, once that's fixed we're good to merge.

TheSin-:
I believe so!!

TheSin-:
Fixed, for me in both views now.

UnchartedBull:
🚢

Issue #1156, opened 2020-11-06
arm64 install issue


Will-wastelander:
**Describe the bug**
Not sure if this is a bug, feature request, or a support request, but I figured I'd mark it as a bug, since it can't be installed via the current installer script.

**To Reproduce**
1. Install OctoPi 0.18.0 RC1 arm64
2. Run OctoDash install script
3. Install script fails, appears to be related to installing octodash_2.1.1_armv7l.deb and not octodash_2.1.1_arm64.deb

**Expected behavior**
As OctoPi becomes aarch64 friendly, the installer script should account for this.

**General Information:**
 - Hardware [Raspberry Pi 4 4GB]
 - OS Info [OctoPi 0.18.0 RC1 arm64]
 - OctoDash Version [e.g. v2.1.1]
 - OctoPrint Version [v1.4.2]

**Additional context**
releaseURL=$(curl -s "https://api.github.com/repos/UnchartedBull/OctoDash/releases/latest" | grep "browser_download_url.*armv7l.deb" | cut -d '"' -f 4)

eliteSchwein:
https://sites.google.com/view/m1dn1ght-3d/downloads/unofficial-octodash-builds?authuser=0

Check this url out

Will-wastelander:
I was able to get this working by adding the below to the install script. I'm not the best at writing scripts, but I can try to piece something together to make it work.

I'm not sure if it's the best way to do it, but it works for me.

```
-releaseURL=$(curl -s "https://api.github.com/repos/UnchartedBull/OctoDash/releases/latest" | grep "browser_download_url.*armv7l.deb" | cut -d '"' -f 4)
+arch=$(uname -m)
+if [[ $arch == x86_64 ]]; then
+    echo "x86_64 Architecture"
+    releaseURL=$(curl -s "https://api.github.com/repos/UnchartedBull/OctoDash/releases/latest" | grep "browser_download_url.*amd64.deb" | cut -d '"' -f 4)
+elif [[ $arch == aarch64 ]]; then
+    echo "ARM64 Architecture"
+    releaseURL=$(curl -s "https://api.github.com/repos/UnchartedBull/OctoDash/releases/latest" | grep "browser_download_url.*arm64.deb" | cut -d '"' -f 4)
+elif  [[ $arch == arm* ]]; then
+    echo "ARM Architecture"
[...]

PR #1157, opened 2020-11-06
Update install.sh

Will-wastelander:
Add multi arch support to installer script.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -679,7 +679,7 @@ elif  [[ $arch == arm* ]]; then
     echo "ARM Architecture"
     releaseURL=$(curl -s "https://api.github.com/repos/UnchartedBull/OctoDash/releases/latest" | grep "browser_download_url.*armv7l.deb" | cut -d '"' -f 4)
 fi
-dependencies="libgtk-3-0 libnotify4 libnss3 libxss1 libxtst6 xdg-utils libatspi2.0-0 libuuid1 libappindicator3-1 libsecret-1-0 xserver-xorg ratpoison x11-xserver-utils xinit libgtk-3-0 bc desktop-file-utils libavahi-compat-libdnssd1"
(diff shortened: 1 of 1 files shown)


UnchartedBull (commented):
looks good. one minor thing maybe:

I think it would be great if you could save the architecture to a string, rather than echoing it directly to the console. Then include that string in the `Installing OctoDash part`. Final output could look something like this:

`Installing OctoDash (v2.1.1, armv7l)`. What do you think about that?

You can ignore the CodeFactor issue btw.

Will-wastelander:
Tested, and this should be working now.

scripts/install.sh
```diff
@@ -723,7 +730,7 @@ if [ $DIRECTORY != "-" ]; then
   fi;
 fi;
 
-echo "Installing OctoDash ..."
+echo "Installing OctoDash "${version[7]}, $arch..."
```
UnchartedBull, line 733:
```suggestion
echo "Installing OctoDash ..."
```

Having the same output line twice might be confusing. After the version and arch has been printed they shouldn't be printed again IMHO.

scripts/install.sh
```diff
@@ -782,7 +789,7 @@ fi
 
 
 list_input "Shall I reboot your Pi now?" yes_no reboot
-echo "OctoDash has been successfully installed! :)"
+echo "OctoDash "${version[7]} $arch" has been successfully installed! :)"
```
UnchartedBull, line 792:
```suggestion
echo "OctoDash has been successfully installed! :)"
```
same as above

UnchartedBull:
Fixes #1156.

scripts/install.sh
Will-wastelander: The only reason I did this, is due to the first version print is scrolled due to package deps. It is not needed, just thought it would be nice to have it shown for each, so you don't have to scroll to find it.

UnchartedBull: Yeah makes sense. I'd still remove the last one then, are you fine with that? Are the `...` printed correctly though? Shouldn't there be another `"` right in front of them?

Will-wastelander: Yea, looks like I missed that. It has been fixed, with the original space that was there. The other line has been reverted also.

UnchartedBull: Thanks! 👍

UnchartedBull (approved):

PR #1256, opened 2020-12-18
Add support for TPLinkSmartPlug plugin

### LINKED ISSUE #1244: Add support for OctoPrint-TPLinkSmartplug as a plugin on OctoDash
_- Firstly, thanks a lot for this project, it is awesome. -_

I personally use [OctoPrint-TPLinkSmartplug](https://plugins.octoprint.org/plugins/tplinksmartplug/) to power on/off my printer, instead of [PSUControl](https://plugins.octoprint.org/plugins/psucontrol/).

I'd like to power off my printer when OctoDash is exiting from sleep mode and, like for PSUControl, add custom actions (like `[!TPLINKON]`, `[!TPLINKOFF]` and `[!TPLINKTOGGLE]`).

**Link other projects**
- OctoPrint-TPLinkSmartplug: https://plugins.octoprint.org/plugins/tplinksmartplug/

**Additional context**
I think it could be straightforward to support it on OctoDash, because the plugin has an API (and a [wiki](https://github.com/jneilliii/OctoPrint-TPLinkSmartplug/wiki/API)), and works like PSUControl (on/off action available (+ status for toggle action)). The main difference is that we have to specify an IP address when we want to change the state of the SmartPlug.

xunleii:
Add the support for TP Link as alternative of PSU Controler plugin (and fix the issue #1244).

It implements 2 custom actions (`[!TPLINKOFF]` and `[!TPLINKON]`) and an option to power on the smartplug when the standby mode exits. I tested these changes on my RPi 3B (and on my laptop) and it seems working without any issues (this is barely a copy/paste of the PSU plugin, without the toggle feature).

Unfortunately, we need a keyboard to setup the IP address in the settings menu (or edit the `config.json` file).

Closes #1244

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -40,6 +40,10 @@ export class AppComponent implements OnInit {
     if (this.configService && this.configService.isInitialized()) {
       if (this.configService.isLoaded()) {
         if (this.configService.isValid()) {
+          // TODO: this call is only here to allow migration path on `turnOnPrinterWhenExitingSleep`
+          //       option release.
+          this.migratePSUControlOption();
+
           this.waitForOctoprint();
         } else {
           this.checkInvalidConfig();
@@ -53,8 +57,10 @@ export class AppComponent implements OnInit {
   }
 
   private checkInvalidConfig() {
-    if (_.isEqual(this.configService.getErrors(), this.service.getUpdateError())) {
-      if (this.service.autoFixError()) {
+    const errors = this.configService.getErrors();
+
+    if (this.service.hasUpdateError(errors)) {
+      if (this.service.autoFixErrors(errors)) {
         this.initialize();
       } else {
         this.configService.setUpdate();
@@ -123,5 +129,17 @@ export class AppComponent implements OnInit {
     //     );
     //   })
     //   .finally(() => clearTimeout(showPrinterConnectedTimeout));
+  }
+
+  // TODO: this method is only here to allow migration path on `turnOnPrinterWhenExitingSleep`
+  //       option release, which can break current PSUControl behaviour.
+  //       It will be removed when all people have used a release with this fix.
+  private migratePSUControlOption() {
+    const config = this.configService.getCurrentConfig();
+    if (config.plugins.psuControl.turnOnPSUWhenExitingSleep) {
+      config.octodash.turnOnPrinterWhenExitingSleep = true;
+      config.plugins.psuControl.turnOnPSUWhenExitingSleep = false;
+      this.configService.saveConfig(config);
+    }
   }
 }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -40,10 +40,6 @@ export class AppComponent implements OnInit {
     if (this.configService && this.configService.isInitialized()) {
       if (this.configService.isLoaded()) {
         if (this.configService.isValid()) {
-          // TODO: this call is only here to allow migration path on `turnOnPrinterWhenExitingSleep`
-          //       option release.
-          this.migratePSUControlOption();
-
           this.waitForOctoprint();
         } else {
           this.checkInvalidConfig();
@@ -60,7 +56,7 @@ export class AppComponent implements OnInit {
     const errors = this.configService.getErrors();
 
     if (this.service.hasUpdateError(errors)) {
-      if (this.service.autoFixErrors(errors)) {
+      if (this.service.fixUpdateErrors(errors)) {
         this.initialize();
       } else {
         this.configService.setUpdate();
@@ -129,17 +125,5 @@ export class AppComponent implements OnInit {
     //     );
     //   })
     //   .finally(() => clearTimeout(showPrinterConnectedTimeout));
-  }
-
-  // TODO: this method is only here to allow migration path on `turnOnPrinterWhenExitingSleep`
-  //       option release, which can break current PSUControl behaviour.
-  //       It will be removed when all people have used a release with this fix.
-  private migratePSUControlOption() {
-    const config = this.configService.getCurrentConfig();
-    if (config.plugins.psuControl.turnOnPSUWhenExitingSleep) {
-      config.octodash.turnOnPrinterWhenExitingSleep = true;
-      config.plugins.psuControl.turnOnPSUWhenExitingSleep = false;
-      this.configService.saveConfig(config);
-    }
   }
 }
--- a/src/app/app.service.ts
+++ b/src/app/app.service.ts
@@ -35,13 +35,15 @@ export class AppService {
         (config.plugins.tpLinkSmartPlug = { enabled: true, smartPlugIP: '127.0.0.1' }),
       ".octodash should have required property 'previewProgressCircle'": config =>
         (config.octodash.previewProgressCircle = false),
-      ".octodash should have required property 'turnOnPrinterWhenExitingSleep'": config =>
-        (config.octodash.turnOnPrinterWhenExitingSleep = false),
+      ".octodash should have required property 'turnOnPrinterWhenExitingSleep'": config => {
+        config.octodash.turnOnPrinterWhenExitingSleep = config.plugins.psuControl.turnOnPSUWhenExitingSleep ?? false;
+        delete config.plugins.psuControl.turnOnPSUWhenExitingSleep;
+      },
     };
   }
 
-  // If the errors can be automatically fixed return true here
-  public autoFixErrors(errors: string[]): boolean {
+  // If all errors can be automatically fixed return true here
+  public fixUpdateErrors(errors: string[]): boolean {
     const config = this.configService.getCurrentConfig();
 
     let fullyAutofixed = true;
--- a/src/app/app.service.ts
+++ b/src/app/app.service.ts
@@ -1,7 +1,9 @@
 import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
+import _ from 'lodash';
 import { ElectronService } from 'ngx-electron';
 
+import { Config } from './config/config.model';
 import { ConfigService } from './config/config.service';
 import { NotificationService } from './notification/notification.service';
 
@@ -9,7 +11,7 @@ import { NotificationService } from './notification/notification.service';
   providedIn: 'root',
 })
 export class AppService {
-  private updateError: string[];
+  private updateError: Record<string, (config: Config) => void>;
   private loadedFile = false;
   public version: string;
   public latestVersion: string;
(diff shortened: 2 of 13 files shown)


UnchartedBull:
Thank you for your contribution! This looks really promising. I'm kind of busy right now, so it probably will take me a day or two to review this. But from a first glance everything looked alright!

src/app/config/config.default.ts
```diff
@@ -48,6 +48,11 @@ export const defaultConfig: Config = {
       enabled: false,
       turnOnPSUWhenExitingSleep: false,
     },
+    tplinkSmartPlug: {
+      enabled: false,
+      smartPlugIP: '127.0.0.1',
+      turnOnPowerWhenExitingSleep: false,
```
UnchartedBull, line 54:
```suggestion
```

I think we shouldn't have this value twice in the config (`turnOnPSUWhenExitingSleep` already exists)

src/app/config/config.model.ts
```diff
@@ -70,6 +71,11 @@ interface PSUControlPlugin extends Plugin {
   turnOnPSUWhenExitingSleep: boolean;
 }
 
+interface TPLinkSmartPlugPlugin extends Plugin {
+  smartPlugIP: string;
+  turnOnPowerWhenExitingSleep: boolean;
```
UnchartedBull, line 76:
```suggestion
```

same as above

src/app/config/config.schema.ts
```diff
@@ -198,6 +198,25 @@ export const configSchema = {
             },
           },
         },
+        tpLinkSmartPlug: {
+          $id: '#/properties/plugins/properties/tplinkSmartPlug',
+          type: 'object',
+          required: ['enabled', 'smartPlugIP', 'turnOnPowerWhenExitingSleep'],
```
UnchartedBull, line 204:
```suggestion
          required: ['enabled', 'smartPlugIP'],
```

src/app/config/config.schema.ts
```diff
@@ -198,6 +198,25 @@ export const configSchema = {
...
+              type: 'boolean',
+            },
+            smartPlugIP: {
+              $id: '#/properties/plugins/properties/smartPlugIP',
+              type: 'string',
+            },
+            turnOnPowerWhenExitingSleep: {
+              $id: '#/properties/plugins/properties/turnOnPowerWhenExitingSleep',
+              type: 'boolean',
+            },
```
UnchartedBull, line 217:
```suggestion
```

src/app/config/config.service.ts
```diff
@@ -197,6 +197,14 @@ export class ConfigService {
...
   }
 
+  public getSmartPlugIP(): string {
+    return this.config.plugins.tplinkSmartPlug.smartPlugIP;
+  }
+
+  public turnOnPowerWhenExitingSleep(): boolean {
+    return this.config.plugins.tplinkSmartPlug.turnOnPowerWhenExitingSleep;
+  }
+
```
UnchartedBull, line 207:
```suggestion
  public useTpLinkSmartPlug(): string {
    return this.config.plugins.tplinkSmartPlug.enabled;
  }

  public getSmartPlugIP(): string {
    return this.config.plugins.tplinkSmartPlug.smartPlugIP;
  }
```

src/app/settings/settings.component.html
```diff
@@ -567,6 +567,43 @@
...
+          >
+            <span class="settings__checkbox">
+              <span
+                class="settings__checkbox-checked"
+                [ngClass]="{ 'settings__checkbox-checked-disabled': !config.plugins.tplinkSmartPlug.enabled }"
+                *ngIf="config.plugins.tplinkSmartPlug.turnOnPowerWhenExitingSleep"
+              ></span>
+            </span>
+            <span class="settings__checkbox-descriptor">Turn on TPLink when exiting sleep</span>
+          </div>
```
UnchartedBull, line 595:
```suggestion
```

src/app/settings/settings.component.html
```diff
@@ -567,6 +567,43 @@
...
+          <label for="smart-plug-ip" class="settings__input-label">Smart plug IP</label>
```
UnchartedBull, line 596:
```suggestion
          <label for="smart-plug-ip" class="settings__input-label">SmartPlug IP</label>
```

I think this might be good for consistent naming (same as heading).

src/app/standby/standby.component.ts
```diff
@@ -40,6 +42,9 @@ export class StandbyComponent implements OnInit {
     if (this.configService.turnOnPSUWhenExitingSleep()) {
       this.psuControlService.changePSUState(true);
       setTimeout(this.checkConnection.bind(this), 5000);
+    } else if (this.configService.turnOnPowerWhenExitingSleep()) {
+      this.tpLinkSmartPlugService.changePowerState(true);
+      setTimeout(this.checkConnection.bind(this), 5000);
```
UnchartedBull, line 47:
```suggestion
    if (this.configService.turnOnPSUWhenExitingSleep()) {
      if (this.configService.useTpLinkSmartPlug()) {
         this.tpLinkSmartPlugService.changePowerState(true);
      } else {
        this.psuControlService.changePSUState(true);      
      }
      setTimeout(this.checkConnection.bind(this), 5000);
```

src/app/settings/settings.component.html
UnchartedBull, line 596:
```suggestion
          <label for="smart-plug-ip" class="settings__input-label">SmartPlug IP</label>
```

I think this might be good for consistent naming (same as heading).

UnchartedBull (changes requested):
Really solid PR for a highly requested feature. Thank you so much!

I just have one really minor issue with this, with your changes the turnOnPowerWhenExitingSleep will exist twice which might be a bit confusing to some users. I'd prefer if we keep the already existing attribute and then just check whether TPLink SmartPlug Plugin is enabled (then use the TPLink SmartPlug) or not (then use PSUControl). I proposed some changes (altough not tested), but they should work I think. Let me know if you have any objections on removing `turnOnPowerWhenExitingSleep` from the TPLink Plugin :)

xunleii:
> Really solid PR for a highly requested feature. Thank you so much!
> 
> I just have one really minor issue with this, with your changes the turnOnPowerWhenExitingSleep will exist twice which might be a bit confusing to some users. I'd prefer if we keep the already existing attribute and then just check whether TPLink SmartPlug Plugin is enabled (then use the TPLink SmartPlug) or not (then use PSUControl). I proposed some changes (altough not tested), but they should work I think. Let me know if you have any objections on removing `turnOnPowerWhenExitingSleep` from the TPLink Plugin :)
> 
> edit: Might be a good idea to rename the toggle to turn on power when exiting sleep (on the UI side should be more than enough) to make it less confusing.

Yep, it's look better to have only one button to do that. Moreover, it allows us to extends OctoDash with another PSU plugin more easier in the future ^^. I will apply these changes and test it today. 

Also, I have an idea for this option; What about moving the option `turnOnPSUWhenExitingSleep` directly in the plugin settings (not affiliated to a specific plugin like `PSUControl`) ? I think it will be less confusing because, when we use the TPLink plugin, we need to toggle an option on the `PSUControl` settings.

Thanks a lot for your review.

xunleii:
After testing with suggestion, moving the `turnOnPSUWhenExitingSleep` option as affiliated options make more sense; currently, we need to enable `PSUControl` plugin to toggle `turnOnPSUWhenExitingSleep` option.

xunleii:
I have another issue ... when I upgrade from the last release to the next with my PR, the settings menu is broken. The only way I found to fix it is to add manually the TPLink plugin settings section inside the `config.json` file.

src/app/app.service.ts
```diff
@@ -25,19 +27,33 @@ export class AppService {
...
+      ".octodash should have required property 'previewProgressCircle'": config =>
+        (config.octodash.previewProgressCircle = false),
+      ".octodash should have required property 'turnOnPrinterWhenExitingSleep'": config =>
+        (config.octodash.turnOnPrinterWhenExitingSleep = false),
+    };
   }
 
   // If the errors can be automatically fixed return true here
-  public autoFixError(): boolean {
+  public autoFixErrors(errors: string[]): boolean {
```
UnchartedBull, line 44: feel free to call this function migrate, since it is a migrate function now rather than just blindly fixing errors.

src/app/config/config.model.ts
```diff
@@ -67,15 +68,22 @@ interface EnclosurePlugin extends Plugin {
 }
 
 interface PSUControlPlugin extends Plugin {
+  // TODO: this option still exists to allow migration path... need to be removed
+  // when the new `turnOnPSUWhenExitingSleep` will be released
   turnOnPSUWhenExitingSleep: boolean;
```
UnchartedBull, line 73:
```suggestion
  turnOnPSUWhenExitingSleep?: boolean;
```

we should mark it optional though when it should get removed in the next release :)

UnchartedBull (commented):
That migrate function is awesome! That is something that I've been wanting to do from the beginning but didn't really know how to tackle. Thanks for cleaning that up and making future config changes a lot easier!

I feel bad to request some more changes, but I think the migratePSUControl is better fitted in your new awesome migrate function! After that this is ready to be merged and will then be included in the next release :)

src/app/app.component.ts
```diff
@@ -40,6 +40,10 @@ export class AppComponent implements OnInit {
     if (this.configService && this.configService.isInitialized()) {
       if (this.configService.isLoaded()) {
         if (this.configService.isValid()) {
+          // TODO: this call is only here to allow migration path on `turnOnPrinterWhenExitingSleep`
+          //       option release.
+          this.migratePSUControlOption();
+
```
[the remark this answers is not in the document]
xunleii: Yep, looks better (I made this function before making changes on the autofix function ^^')

src/app/app.service.ts
```diff
@@ -25,19 +27,33 @@ export class AppService {
...
+
+    // list of all error following an upgrade
+    this.updateError = {
+      ".printer should have required property 'zBabystepGCode'": config => (config.printer.zBabystepGCode = 'M290 Z'),
+      ".plugins should have required property 'tpLinkSmartPlug'": config =>
+        (config.plugins.tpLinkSmartPlug = { enabled: true, smartPlugIP: '127.0.0.1' }),
+      ".octodash should have required property 'previewProgressCircle'": config =>
+        (config.octodash.previewProgressCircle = false),
+      ".octodash should have required property 'turnOnPrinterWhenExitingSleep'": config =>
+        (config.octodash.turnOnPrinterWhenExitingSleep = false),
```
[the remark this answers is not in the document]
xunleii: I will try that today. If the second suggestion works, it would be awesome.

src/app/app.service.ts
xunleii: What about `fixMigrationErrors` ?

UnchartedBull: You made the function, I'm fine with either, so choose the one you like better :)

xunleii: Thanks :)

src/app/app.service.ts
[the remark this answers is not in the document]
xunleii: The suggestion works. I just added the terser operator to avoid mistakes (and force `false` even if the field is null)

xunleii:
All is fixed. I also use the [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) in [`getErrors`](https://github.com/UnchartedBull/OctoDash/pull/1256/files#diff-07e4ab09e4786501cbe029d146c0b73165e221a55dbb3ba5bc6097c3250e00bcR83) to avoid null exception when no error are found (and update the [`valid`](https://github.com/UnchartedBull/OctoDash/pull/1256/files#diff-07e4ab09e4786501cbe029d146c0b73165e221a55dbb3ba5bc6097c3250e00bcR100) field when the configuration is updated).

Thanks for your review and your suggestions.

UnchartedBull (approved):
🚢

Jack77777777:
Can be TPLink Plug used to turn on/off printer through Octodash and power on when exiting stand by ?
I installed Octodash Today but can´t find how to configure TPLink PLugin

PR #1258, opened 2020-12-20
Add support of Basic Auth for OctoPrint connection

anthonymougin:
Hi,

This is my first contribution to your great project, so i would like to primary say thanks a lot !
And i'm sorry for my bad English, i'm French ;).

My OctoPrint server use HTTP Basic Auth and i was not able to connect to it.

I write a new Setup Page just after the Connection Url page, where the user can enter a user and pass to add Authorization Headers to all octoprint request. I moved all Headers generation to the configService, so we can't missed any headers anymore.


anthonymougin:
I will fix lint errors today

UnchartedBull:
Thank you for your contribution! These are a lot of changes, so it might take me some time to get through them, plus I didn't even knew that OctoPrint supported BasicAuth for the API. I'll make sure to review everything before the next release though! And thanks for already fixing the linting issues :)

src/app/config/config.default.ts
```diff
@@ -2,7 +2,7 @@ import { Config } from './config.model';
 
 export const defaultConfig: Config = {
   octoprint: {
-    url: 'http://localhost:5000/api/',
+    url: 'http://localhost:3000',
```
UnchartedBull, line 5: Why are you changing the default here? The /api/ is still required and port 3000 is not the default port.

src/app/config/config.model.ts
```diff
@@ -12,12 +12,18 @@ export interface Config {
...
+export interface Octoprint {
   url: string;
   accessToken: string;
+  basicAuth?: BasicAuth;
   urlSplit?: URLSplit;
 }
 
+export interface BasicAuth {
+  user?: string;
+  pass?: string;
```
UnchartedBull, line 24: If those are required in the Schema they shouldn't be optional here either
```suggestion
  user: string;
  pass: string;
```

src/app/config/config.service.ts
```diff
@@ -47,14 +47,24 @@ export class ConfigService {
     this.initialized = true;
   }
 
+  public generateHttpHeadersForConfig(octoprint: Octoprint): HttpHeaders {
```
UnchartedBull, line 50:
```suggestion
  public generateHttpHeadersForGivenInstance(octoprint: Octoprint): HttpHeaders {
```

Might make this a little clearer :)

src/app/config/setup/octoprint-authentication/octoprint-authentication.component.html
```diff
@@ -2,6 +2,7 @@
   Please authenticate me now. You can either click the button below and confirm the request in your OctoPrint
   webinterface or enter an API Key manually.
 </span>
+<div></div>
```
UnchartedBull, line 5: why is there an empty div here?

src/app/config/setup/octoprint-basic-auth/octoprint-basic-auth.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  set user(user: string) {
+    if (!this.basicAuth) {
+      this.basicAuth = {};
+    }
+    this.basicAuth.user = user;
+  }
+
+  get pass(): string {
+    return this.basicAuth && this.basicAuth.pass;
```
UnchartedBull, line 26: Doesn't that just return true or false, based on whether the objects are available? Another guy showed me https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining in a PR the other day, might be something worth considering here (if it makes sense).

src/app/config/setup/personalization/personalization.component.ts
```diff
@@ -11,19 +12,21 @@ export class PersonalizationComponent implements OnInit {
...
 
   ngOnInit(): void {
-    this.printerProfileService.getActivePrinterProfileName(this.octoprintURL, this.apiKey).subscribe(printerName => {
+    this.printerProfileService.getActivePrinterProfileName(this.octoprintURL, this.apiKey, this.basicAuth).subscribe(printerName => {
       if (!this.printerName) {
         this.printerName = printerName;
         this.printerNameChange.emit(this.printerName);
       }
     });
+    console.log(this.basicAuth);
```
UnchartedBull, line 29: please remove
```suggestion
```

src/app/octoprint/auth.service.ts
```diff
@@ -1,24 +1,33 @@
...
 export class AuthService {
   constructor(private http: HttpClient) {}
 
-  public probeSupport(octoprintURL: string): Observable<HttpResponseBase> {
-    return this.http.get<void>(`${octoprintURL}plugin/appkeys/probe`, { observe: 'response' });
+  private basicAuthHeaders(basicAuth?: BasicAuth) {
+    if (basicAuth) {
+      return new HttpHeaders({
+        'Authorization': 'Basic ' + btoa(`${basicAuth.user}:${basicAuth.pass}`)
+      });
```
UnchartedBull, line 17: I think just returning empty headers here would be nice.

src/app/octoprint/auth.service.ts
```diff
@@ -1,24 +1,33 @@
...
 
 @Injectable({
   providedIn: 'root',
 })
 export class AuthService {
   constructor(private http: HttpClient) {}
 
-  public probeSupport(octoprintURL: string): Observable<HttpResponseBase> {
-    return this.http.get<void>(`${octoprintURL}plugin/appkeys/probe`, { observe: 'response' });
+  private basicAuthHeaders(basicAuth?: BasicAuth) {
```
UnchartedBull, line 13: please add function return type

src/app/settings/settings.component.html
```diff
@@ -66,6 +66,41 @@
...
+                            name="basicAuthUser"
+                            style="width: 19vw"
+                            [(ngModel)]="config.octoprint.basicAuth.user"
+                            required
+                    />
+                  </td>
+                  <td>
+                    <label for="basicAuthPass" class="settings__input-label">Pass</label>
+                    <input
+                            type="text"
```
UnchartedBull, line 93:
```suggestion
                            type="password"
```

should be type password, otherwise anyone could read it by just walking up to the Pi

src/app/settings/settings.component.ts
```diff
@@ -41,6 +43,8 @@ export class SettingsComponent implements OnInit {
   ) {
     this.config = this.configService.getCurrentConfig();
     this.config.octoprint.urlSplit = this.configService.splitOctoprintURL(this.config.octoprint.url);
+
+    this.useBasicAuth = !!this.config.octoprint.basicAuth;
```
UnchartedBull, line 47: why is there a double `!!` here?

UnchartedBull (changes requested):
Thanks again for your contribution. I don't know anything about Basic Auth and Octoprint really so please excuse some (probably) stupid questions. There are some things that need addressing though:

- If I select "enter manually" and DON'T select Use Basic Auth I'm still shown the Basic Auth input mask (once I enable and disable it is correctly hidden)
- The API Key screen is always being shown, I think it should be hidden if BasicAuth is being used?
- A little more complicated bug:
  - Tick Use Basic Auth
  - Enter Username
  - Delete Username and go back
  - Disable Basic Auth
  - Config will not be valid and will be rejected on the final screen: `[".octoprint.basicAuth should be object"]`
  - the console.log switches from undefined to null which trips some logic I guess
- I'm not 100% sure whether I like the placement of the Basic Auth option, I think it is more fitting in the API Key section (could be integrated similar to how automatic discovery / manual IP works). This would also remove the need to create another page, which needs to be hidden in some cases, so it would just be a single page, that switches between two layouts.
- Plus some minor remarks in the code

Really great contribution, just needs some polishing and bug fixing and then it should be good to go :)

UnchartedBull:
Oh wait, is the intended use case for this to authenticate a Proxy that is before the actual OctoPrint installation?

UnchartedBull:
I just stumbled upon almost the exact same issue on PrusaSlicer (issue 650) and they mentioned that something like: http://username:password@example.com/ might also work. I'm not 100% sure whether the angular http library does support that but it might be worth a shot. Still the headers are the better and cleaner solution IMHO. But the issues lined out in the PrusaSlicer thread apply here as well, espacially:

> We would also need to explain to other users that this login+password is not the one used for logging into Octoprint and that it's something completely else which most of them will probably never need. I'm not sure how this could be done without confusing people.

There already is a lot of confusion regarding some of the config entries and I can imagine a few popping up here and there asking why OctoDash isn't working even though they entered their OctoPrint credentials in the BasicAuth part (which obviously doesn't work).

Long story short: I'm really liking your contribution as it's improving a lot of other places as well plus the Basic Auth is useful, I'd just remove the UI part. Anyone who is able to setup a reverse proxy with BasicAuth protection should be able to edit a JSON file and add two lines there. What do you think about that? Of course there will be a wiki entry that outlines the procedure. There are a few options by now which are "manual edit only" to simplify setup & settings as much as possible and I think this one fits in there as well.

anthonymougin:
> Long story short: I'm really liking your contribution as it's improving a lot of other places as well plus the Basic Auth is useful, I'd just remove the UI part. Anyone who is able to setup a reverse proxy with BasicAuth protection should be able to edit a JSON file and add two lines there. What do you think about that? Of course there will be a wiki entry that outlines the procedure. There are a few options by now which are "manual edit only" to simplify setup & settings as much as possible and I think this one fits in there as well.

I think you'r right. Maybe this feature will be confusing for "basic" OctoPrint user because this is not a OctoPrint config but a server one.

Let's do that : Remove UI config and explain in the wiki how to modify the JSON to enable the Basic Auth header.

UnchartedBull:
That sounds like a solid plan. Will you update the PR?

PR #1266, opened 2020-12-30
Wifi Menu

WalnutFox:
This is a wifi menu that i have added to octodash, i can only confirm that it works in the dev environment. i will update this when i get a chance to test on my rpi. 

the only problem with this at the moment is that it requires root access and a keyboard.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1686,7 +1686,6 @@
       "version": "1.12.2",
       "resolved": "https://registry.npmjs.org/@electron/get/-/get-1.12.2.tgz",
       "integrity": "sha512-vAuHUbfvBQpYTJ5wB7uVIDq5c/Ry0fiTBMs7lnEYAo/qXXppIVcWdfBr57u6eRnKdVso7KSiH6p/LbQAG6Izrg==",
-      "dev": true,
       "requires": {
         "debug": "^4.1.1",
         "env-paths": "^2.2.0",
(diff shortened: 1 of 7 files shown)


WalnutFox:
for this to work without password prompt i recommend setting iwlist, wpa_cli and rfkill are set to nopasswd in the sudoers file.

WalnutFox:
I have confirmed that this does indeed work on the raspberry pi, I will be adding a bit more functionality to the wifi menu as well to tell you the ip address.

WalnutFox:
I have added IP and SSID to the wifi-menu that gets updated after connecting. added dhclient release and renew to allow the pi to get a new ip address.

recommended to add dhclient to visudo as well.

WalnutFox:
If anyone knows how to setup popups to let user know delay reason, let me know. next commit is my attempt but it is missing stuff to get it to work.

UnchartedBull:
Thanks for you contribution! :) Just had a quick brush over this:

- If I don't enable sudo it would just hang with no way of exiting other than entering the sudo password on the CLI
- Not entirely a fan of the CLI usage in the actual frontend
- Doesn't work an a Mac (not a huge issue, but makes it a little harder to test for me at least)

Here are my 2 cents:

- It would be great if all the stuff that needs node and / or CLI commands could be pushed over to the electron side of things
- I stumbled upon these two libraries: https://www.npmjs.com/package/node-wifi and https://www.npmjs.com/package/ip which do seem to achieve the exact same thing without the need for sudo (at least from what I can tell from their READMEs)

I know this is asking quite a lot (and I'm happy to help here if you want me to), but it would be great if we could try the mentioned libraries and check whether they may be able to achieve the functionality without needing privileged rights. They should be encapsulated in their own wrapper similar to how the bonjour implementation is done (https://github.com/UnchartedBull/OctoDash/blob/master/helper/discover.js). Communication would be done via the IPC, this would also keep the logic encapsulated in its own component and away from the settings component so it could be easily reused in other screens.

UnchartedBull:
Oh and if you want to send a notification use the NotificationService :)

WalnutFox:
i'll be honest i dont really have a good understanding of using angular or electron and most of the code is borrowed then change to get it to work. the original source for the code is mozilla gateway for iot. 

with my current knowledge i dont know how to keep things separate between angular and electron.

i do know that on linux however you cannot run those command and use them without using sudo in front of them.

PR #1359, opened 2021-01-30
Add klipper support for changing filament

DS-Ale:
Hi, it's a simple implementation to allow custom gcode to disable the extruder. Using Klipper, M18 would disable all steppers, ignoring arguments.
At first I had added a "klipper support" checkbox into Settings, however I then chose to implement it just like the babystepping option, for the sake of consistency and flexibility. So I added a 'disableExtruderGCode' key, changable in the config.json (default value is 'M18 E'). For Klipper I use "SET_STEPPER_ENABLE STEPPER=extruder ENABLE=0", I can confirm it works (tried @pause and M0/M1).

Just a note. I had to downgrade electron-store to 6 (can't read config.json otherwise) and electron to 9 ("tilted" display otherwise) to get interface working (even before my changes, on clean setup). Of course I reverted dependencies to original values before commit. I don't know if these are known issues or I did something wrong.
This is my first PR too, hope I did everything right.

### DIFF
--- a/src/app/config/config.default.ts
+++ b/src/app/config/config.default.ts
@@ -9,6 +9,7 @@ export const defaultConfig: Config = {
     name: '',
     xySpeed: 150,
     zSpeed: 5,
+    disableExtruderGCode: 'M18 E',
     zBabystepGCode: 'M290 Z',
     defaultTemperatureFanSpeed: {
       hotend: 200,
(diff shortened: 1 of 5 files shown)


UnchartedBull:
The versions are currently a known issue unfortunately. PR looks good from my side, only config update is missing, but I can fix those. Thanks for the changes!

UnchartedBull (approved):

PR #1581, opened 2021-03-20
feature/tasmota

jneilliii:
* add Tasmota plugin power options
* add TasmotaMQTT plugin power options


jneilliii:
@UnchartedBull I'm not sure the best way to handle the initial setup wizard for the config, but during my testing the last Tasmota MQTT option for installed plugins appears to be out of view and it wasn't automatically scrollable (at least in windows dev testing).

UnchartedBull:
I'll have a look at this after v3 beta is released, should be included in the final v3 release! :)

src/app/app.service.ts
```diff
@@ -31,6 +31,10 @@ export class AppService {
       ".printer should have required property 'zBabystepGCode'": config => (config.printer.zBabystepGCode = 'M290 Z'),
       ".plugins should have required property 'tpLinkSmartPlug'": config =>
         (config.plugins.tpLinkSmartPlug = { enabled: true, smartPlugIP: '127.0.0.1' }),
+      ".plugins should have required property 'tasmota'": config =>
+        (config.plugins.tasmota = { enabled: true, ip: '127.0.0.1', idx: null }),
+      ".plugins should have required property 'tasmotaMqtt'": config =>
+        (config.plugins.tasmotaMqtt = { enabled: true, topic: 'topic', relayN: null }),
```
UnchartedBull, line 37: I must have missed that on the Tp-Link SmartPlug PR, the default value for enabled should be `false` here. Would be great if you could also change that for the Line 33.

src/app/config/config.model.ts
```diff
@@ -78,6 +80,16 @@ interface TPLinkSmartPlugPlugin extends Plugin {
   smartPlugIP: string;
 }
 
+interface TasmotaPlugin extends Plugin {
+  ip: string;
+  idx: number;
```
UnchartedBull, line 85: Is `idx` index? Might be beneficial to have `index` as the attribute name instead to prevent confusions.

src/app/config/config.model.ts
```diff
@@ -78,6 +80,16 @@ interface TPLinkSmartPlugPlugin extends Plugin {
...
+}
+
+interface TasmotaMqttPlugin extends Plugin {
+  topic: string;
+  relayN: number;
```
UnchartedBull, line 90: I think a more describing attribute name would also be a better fit here. If its the relay number `relayNumber` might be a good option.

UnchartedBull (changes requested):
Solid PR. Thank you! Would be great if you could replace `idx` and `relayN` with more descriptive variable names (plus the functions in the configService). If that is done this is ready to be merged. I'll fix the initial setup before the final release then.

src/app/config/config.model.ts
jneilliii: idx changed to index throughout

jneilliii:
requested changes completed

UnchartedBull (approved):

UnchartedBull:
Thanks!

Issue #1622, opened 2021-03-28
progressive rewrite of hardcoded vw values to dynamic flex values


pciavald:
## is your feature request related to a problem? Please describe.

I've noticed that a lot of the interface uses hardcoded values to define paddings, margins and positioning. You've noted in another issues that there was often issues opened by users about unexpected layouts, and I would tend to think that there's a link between these two. Also, contributing to and adjusting the interface is not easy when every distance and position is hardcoded.

## Describe the solution you'd like

Progressively transition updated components to [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/). Flexbox is a native css feature that allows dynamic setting of margins, sizes, positions and page flows.

## Additional context

This issue is only intended to be closed with a positive or negative answer. Subsequent issues will be opened for any modified component.

#### before flexbox: 
- visual, manual, hardcoded adjustment of the red boxes margins to center the arrows
- distance selection not actually centered

![image](https://user-images.githubusercontent.com/3384097/112758333-160ef700-8fee-11eb-835d-40abbb6381ef.png)

#### after flexbox:
- cleaner content handling
- boxes are arithmetically centered and will flow depending on the width and height
- an additional jog control can be added without changing any margins, paddings or widths (no css modified)

![image](https://user-images.githubusercontent.com/3384097/112758225-6df92e00-8fed-11eb-922e-265b41038d94.png)

![image](https://user-images.githubusercontent.com/3384097/112758403-874eaa00-8fee-11eb-9828-30ba171d7250.png)

UnchartedBull:
I think you understood me wrong there. The issues a couple of users have reported are due to a completely wrong resolution - i.e. OctoDash is rendering in 1920x1080, and the screen is displaying only the top left 800x480 pixels.

Nevertheless flexboxes are super nice, I haven't really worked with them much when I started OctoDash, which is why I used vw and hw (which are at least somewhat dynamic compared to px units). 

So it would definitely be nice to use more flexboxes, I wouldn't put this at high priority though. If someone needs to do UI changes to a component anyways (like you to the control screen) it would be nice to replace the existing layout with flexboxes, but I think we don't need to replace everything with flexboxes just because now. It might not be the cleanest way but I haven't heard of any issues regarding the actual layout rendering wrong so far and there are still a lot of other open issues :)

pciavald:
Ok perfect, exactly what i thought too. Can you point out to an issue opened by a user of what you're talking about? i can take a look.

PR #1623, opened 2021-03-28
E movement

### LINKED ISSUE #1591: Add control for E axis

We need to be able to move the E axis from the dashboard, which is the extruder motor moving the filament.

We can implement in the Control screen a second vertical jog like for Z, but for E. Should it be made in the component directly or in an additional module ? (i did not check the modules structure yet, maybe it's not possible).

### LINKED ISSUE #1883: Controll manual extrusion - "feature request"
Next to controlling x, y an z-axis it would be nice if extrusion could be manual controlled.
Easy to clean hot-end when swapping color and also when checking e-steps.
Icon buttons could be next to Z-axis buttons.

pciavald:
This is how we implement it for us. I needed to refactor to flexbox (#1622) for easier enabling/disabling of the component. Feel free to just close if you don't want it in upstream, or use only the flexbox refactor.

- implement new configuration `showExtruderControl` defaulting to false
- control extrusion in control component if showExtruderControl set to true in user config

image

Fixes #1591
Closes #1883


pciavald:
It all works, but i still see this error clearly related to my usage of the configService, i guess i missed something to load the service properly ?

image

UnchartedBull:
Great idea to make this configurable! Then there is nothing in the way of merging this into main. Just two little things:

- If you add new config options please make sure to also include them here: https://github.com/UnchartedBull/OctoDash/blob/main/src/app/app.service.ts#L30. This will then automatically upgrade the config if a user upgrades OctoDash. I think in this case it would definitely make sense to add this attribute there and set it to false by default.
- I think it might be nice to replace the icons, this would make it easier to differentiate between Z-Axis Control and Extruder Control without having to read the text. What do you think of `undo-alt` (https://fontawesome.com/icons/undo-alt?style=solid) for retracting and `redo/alt`(https://fontawesome.com/icons/redo-alt?style=solid) for extruding?

The config error should only show up if the devTools reload your app. The problem here is that Angular somehow skips the initialisation process if a different view than mainView is open. This shouldn't happen on the Pi, since the UI is not being reloaded there. So nothing to worry about :)

pciavald:
Thanks, I'll check that! Glad you like the idea of making it configurable!

About the new config i was thinking to let it optional and not include it in the config, only setting the default value to false. This way when a config does not have the property, it does not get in the way, otherwise it can be set to false (no effect) or true. Do you prefer adding them to false anyway ?

UnchartedBull:
Oh I see how this is working now. That's definitely a creative way to exploit the lazy type checking of JS :D.

Would be great if this could be a required part of the config, just to make everything a bit more consistent (should only be two lines of code to add this attribute during upgrades). Thanks! :)

pciavald:
will do!

pciavald:
I've just noticed that this may break themes.

pciavald:
note to self: pushed this branch to upstream to base #1895 on it

PR #1661, opened 2021-04-08
action command prompts

### LINKED ISSUE #1624: action commands support

There's a gcode feature called action commands. It's used to prompt the user when something happens with the printer, for instance a filament runout is detected, or a filament change is going on and the printer wants to know if the color is correct, or a simple message like "bed leveling done".

Similar to how octoprint implemented it (and i guess their plugin can be useful), we should implement a machine prompt modal.

**Link other projects**

https://reprap.org/wiki/G-code#M876:_Dialog_handling

https://docs.octoprint.org/en/master/features/action_commands.html

https://docs.octoprint.org/en/master/bundledplugins/action_command_prompt.html

```
<<< //action:prompt_begin Filament runout detected. Please choose how to proceed:
<<< //action:prompt_choice Swap filament
<<< //action:prompt_choice Home X/Y and pause print
<<< //action:prompt_choice Abort print
<<< //action:prompt_show
> M876 S1
```

https://docs.octoprint.org/en/master/bundledplugins/action_command_notification.html

```
Recv: //action:prompt_end
Recv: //action:notification Bed leveling done.
```

**Additional context**

image

image

pciavald:
Implements support for action:prompt and action:notification

Fixes #1624


UnchartedBull:
the socket service shouldn't be aware of the promptService. This should rather be handled by the event service and should be pushed through the `eventSubject`. This will then be received by the [event-serve](https://github.com/UnchartedBull/OctoDash/blob/main/src/app/event.service.ts) to which the prompt-component should then subscribe. I think we can then also get rid of the prompt-service, since this should rather be fullscreen error, right?

I didn't had time to check out the notification service error today unfortunately, will do this tomorrow.

pciavald:
fullscreen error sound correct as a base for the prompts, i can add the buttons in there. Did you have a chance to take a look? that is our last blocker, the rest is improvement

pciavald:
I can't see any reference to fullscreen errors in the code, is it something you're working on? Maybe in that case it would be better named fullscreen messsage that can be an error, printer notification or prompt.

UnchartedBull:
Ah no that was more of a question on how stuff from the actions_command_prompt should be displayed. If it completely seizes printer operation in all cases it should be a fullscreen thing (like the sleep screen), otherwise it should be similar to notifications.

In the second case it would also make more sense to extend the notification component, instead of creating a new one.

pciavald:
I think actually it makes sense to extend the notification component in both cases, i'll try it now

pciavald:
@UnchartedBull I don't see how i can pass the prompt through the event service without heavy modification, as it needs at least a title and a list of choices. I will keep the notification service in the socket service until a better way to do it is described.

UnchartedBull:
Create another interface in `event.model.ts` (UserPrompt i.e.), update the Subject to return either PrinterEvent or UserPrompt. Abstract function definition would look like this:

`abstract getEventSubscribable(): Observable<PrinterEvent | UserPrompt>`

Check what type the input is in `event-service` and act accordingly. Ideally this would be in two methods (`handlePrinterEvent` and `handleUserPrompt`), which would be called from the `subscribe` function. Move the logic / call of notification service to the `handleUserPrompt` function. This would then also allow the removal of `prompt-service` all-together and also allows the import of the NotificationService in the SocketService.

pciavald:
i've started from scratch, can you check the new architecture ?

src/app/event.service.ts
```diff
@@ -15,35 +16,89 @@ export class EventService implements OnDestroy {
...
-          }, 1000);
+          this.handlePrinterEvent(event);
         }
       }),
     );
   }
 
+  // https://stackoverflow.com/questions/14425568/interface-type-check-with-typescript
+  // https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript/8511350#8511350
+  private isPrinterNotification(object: any): object is PrinterNotification {
```
UnchartedBull, line 35: Maybe another way of doing this would be to check whether the received message is NOT of type `int` (which would indicate that it's an enum, thus PrinterEvent). It should work (haven't tested it though), but might be a bit cleaner.

I wouldn't switch the logic around though, as the default should be handle PrinterEvent. Alternatively you could check whether the received message is of type `object`.

src/app/event.service.ts
```diff
@@ -15,35 +16,89 @@ export class EventService implements OnDestroy {
...
+        messages[event.text] || event.text
+      );
+    } else {
+      // event is action:notification
+      // this.notificationService.setInfo(
+      //   $localize`:@@printer-information:Printer information`,
+      //   messages[event.message] || event.message
+      // );
+      // TODO: annoying as a notification
+      // should be put with an autoclear timeout in the bottom-right statusline
```
UnchartedBull, line 74: I think it still should be a notification, just to keep everything consistent. If this isn't too important it might be a good idea to introduce a fourth notification type with a white strip and just set a timeout here to clear the notification after 5s (is that enough time?).

What do you think of that?

src/app/notification/notification.component.html
```diff
@@ -1,9 +1,29 @@
...
   (click)="hideNotification()"
 >
   <span class="notification__heading">{{ notification.heading }}</span>
   <span class="notification__text">{{ notification.text }}</span>
   <span class="notification__close" i18n="@@tap-close">tap this card to close it</span>
 </div>
+
+<div
+  class="notification-prompt"
+  *ngIf="notification.choices?.length > 0"
```
UnchartedBull, line 14: Wouldn't it be enough to check whether `notification.choices` exist here?

src/app/notification/notification.component.ts
```diff
@@ -35,6 +44,21 @@ export class NotificationComponent implements OnDestroy {
...
+        { command: 'select', choice: index },
+        this.configService.getHTTPHeaders()
+      )
+      .pipe(
+        catchError(error =>
+          this.notificationService.setError($localize`:@@error-answer-prompt:Can't answer prompt!`, error.message),
+        ),
+      )
+      .subscribe();
+  }
```
UnchartedBull, line 60: this should be in `system-service` (or another service if there's one that's better fitting) since it won't be compatible with Klipper / other interfaces. 

From now on all communication with OctoPrint should be behind an abstract service with a generic interface to enable easy integration of other printer software.

src/app/services/socket/socket.octoprint.service.ts
```diff
@@ -116,6 +116,27 @@ export class OctoPrintSocketService implements SocketService {
...
+        },
+      },
+      {
+        check: (plugin: string) => ['action_command_prompt', 'action_command_notification'].includes(plugin),
+        handler: (data: unknown) => this.eventSubject.next(data as PrinterNotification),
+      },
+    ];
+
+    plugins.forEach(plugin =>
+      plugin.check(pluginMessage.plugin.plugin) && plugin.handler(pluginMessage.plugin.data)
```
UnchartedBull, line 136: That definitely looks complicated (which isn't bad). Just curious: Will call this both check methods from the plugins array? Seems like a nice way to do this, since it's easily extendable 👍

UnchartedBull: Ok I think I understood it. Correct me if I'm wrong:

This will go through top to bottom and call the check method with the plugin string. If the check message returns true it will call the handler in the same object, thus handle the message. Otherwise it will just fall-through until it either matches or reaches the end of the array. Really elegant way of handling this 👍

UnchartedBull (commented):
Yes this looks really good! Definitely the right way to go here I think. There are some minor things, most of them are just me being curious though :D. It also would be great if you could attach a screenshot on how the choices thing looks like :)

Once the last few kinks are ironed out this is ready to be merged. I originally planed on releasing v3 on Friday, could easily postpone to Monday or Tuesday though, if this PR should make it into v3.

src/app/event.service.ts
pciavald: I need to retest what exactly was being sent, but AFAIR it was every heat enable/disable message (not always coherent with the user actions) and all changes of % being broadcasted, really cluttering information. We have been using it for a month without these and it seems on-point.

src/app/notification/notification.component.html
pciavald: this is just in case an empty list is sent, we would not want to display a prompt without choices

pciavald:
I am sorry to announce that I do not have the work power to maintain all our changes in compliance with upstream requirements. This feature has been published to our repo [CosmoDash](https://github.com/Cosmyx/CosmoDash).

pciavald:
The code on this branch was in working state last time I checked so it should be pretty easy to finish up and merge !

PR #1701, opened 2021-04-17
Update install.sh

Will-wastelander:
Enable CORS by default

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -748,8 +748,9 @@ echo $cors
 if [ $cors == 'yes' ]; then
         echo "Enabling CORS ..."
         "$DIRECTORY"/bin/octoprint config set --bool "api.allowCrossOrigin" true
-fi
-fi
+else
+  echo "\e[31mCORS has \e[1mNOT\e[22m been enabled. OctoDash most likely won't work if CORS is disabled. You can always enable it in the API settings in OctoPrint"
+fi;
 
 echo "Installing OctoDash "${version[7]}, $arch" ..."
 cd ~
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -739,8 +739,10 @@ if [ $DIRECTORY != "-" ]; then
   fi;
 fi;
 
-echo "Enabling CORS ..."
-/home/pi/oprint/bin/octoprint config set --bool "api.allowCrossOrigin" true
+if /home/pi/oprint/bin/octoprint config get --yaml "api.allowCrossOrigin" | grep -q 'false'; then
+        echo "Enabling CORS ..."
+        /home/pi/oprint/bin/octoprint config set --bool "api.allowCrossOrigin" true
+fi
 
 echo "Installing OctoDash "${version[7]}, $arch" ..."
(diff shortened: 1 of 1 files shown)


UnchartedBull:
Thank you for this! Could we maybe have a user prompt before enabling CORS? As this is a change to security in someway it might be good to ask the user whether he wants to enable CORS. It should be clear that OctoDash doesn't work without CORS enabled though.

Will-wastelander:
Testing that I did for latest update.

pi@Ender-3-octo:~ $ /home/pi/oprint/bin/octoprint config get --yaml "api.allowCrossOrigin"
false
...

pi@Ender-3-octo:~ $ bash ncheck_CORS.sh
? Should I enable CORS ? FYI, this is required by OctoDash v3, and may have security implications yes
yes
Enabling CORS ...
pi@Ender-3-octo:~ $ /home/pi/oprint/bin/octoprint config get --yaml "api.allowCrossOrigin"
true
...

Will-wastelander:
image

scripts/install.sh
```diff
@@ -739,6 +739,18 @@ if [ $DIRECTORY != "-" ]; then
   fi;
 fi;
 
+if "$DIRECTORY"/bin/octoprint config get --yaml "api.allowCrossOrigin" | grep -q 'false'; then
+yes_no=( 'yes' 'no' )
+
+list_input ${red}${bold}"Should I enable CORS ? FYI, this is required by OctoDash v3 and OctoPrint 1.6.0, and may have security implications"${normal} yes_no cors
```
UnchartedBull, line 745:
```suggestion
list_input "Should I enable CORS ? FYI, this is required by OctoDash v3 and OctoPrint 1.6.0, and may have security implications"${normal} yes_no cors
```

I would prefer this not being red, as this is nothing bad (yet).

scripts/install.sh
```diff
@@ -739,6 +739,18 @@ if [ $DIRECTORY != "-" ]; then
...
+
+echo $cors
+if [ $cors == 'yes' ]; then
+        echo "Enabling CORS ..."
+        "$DIRECTORY"/bin/octoprint config set --bool "api.allowCrossOrigin" true
+fi
+fi
```
UnchartedBull, line 752:
```suggestion
else
  echo "\e[31mCORS has \e[1mNOT\e[22m been enabled. OctoDash most likely won't work if CORS is disabled. You can always enable it in the API settings in OctoPrint"
fi;
```

and rather print a big warning if the user chooses no.

UnchartedBull (commented):
Looks good, I'd rather print the error only if the user selects no though. I haven't tested the proposed changes, but they should work. After this is ready to be merged.

Thanks for all your work! :)

Will-wastelander:
The color codes didn't work. Looks like you want "CORS is NOT" in red, with "NOT" being bold, correct ?

image

Will-wastelander:
This is the best I got. 
image

UnchartedBull (approved):

UnchartedBull:
Thanks! :)

PR #1759, opened 2021-05-05
[WIP] packaging, service, upgrade changes

kantlivelong:
This PR changes a chunk of things but they all kind of tie into each other.

### Changelog / Notes
- Rework deb packaging using native deb tooling instead of electron-builder/fpm which has limited options.

- Make use of docker for builds. This is to provide debian package tooling on non debian systems. 

- Rework upgrade process in OctoDash to update itself via file download and extract. No need to install new debs. Follows similar model to [Sonarr](https://github.com/Sonarr/Sonarr). Currently uses rsync but perhaps there's a better way in node.

- Add --version flag to octodash. NOTE: I couldn't figure out how to make it work over a headless session without prepending DISPLAY=:0 to octodash. (Ex `DISPLAY=:0 octodash --version`) It also throws some weird error regardless but shows the version. Perhaps you know of a better method or a fix. Not truly required but seems like it's worth having.

- Use a systemd service to launch octodash with configurable user. Create user if not exist. (Defaults to: pi)

- Remove ratpoison as it doesn't appear to be necessary.

- Drop the `upgrade.sh` script as it's not needed.

- Update the install/remove scripts. Might be completely removable if OctoDash no longer needs to install plugins.

- Make use of environment config files to use test URLs for dev work. (releaseCheckUrl)

### Build Instructions:
Additional requirements: make, docker(non-root)
1. Checkout
2. Run `make`
3. Packages will build and be dumped into the `package/` directory as a sub directory with a unix timestamp.

### Install Instructions:
1. Download appropriate deb
2. `apt-get -f install ./DEBFILE.deb`
3. Reboot or `systemctl start octodash`

### Upgrade Instructions (Option A):
1. Update using the deb manually.
2. Reboot or `systemctl start octodash`

### Upgrade Instructions (Option B):
1. OctoDash will prompt for update.
2. User selects upgrade as usual
3. OctoDash downloads appropriate tar file and extracts to a tempdir.
4. OctoDash synchronizes contents of tempdir to `/opt/OctoDash`. Extra/orphan files will be deleted.
5. OctoDash restarts
NOTE: The deb does not to be updated at this point unless changes need to be made outside of the `/tmp/OctoDash` directory.

### Comments / Questions
1. I'm sure there are some fixes/changes that will be needed.
2. I'm not a super proficient w/ deb packaging tooling but based on what I've done before and how this project works I think this will work. I'm sure someone out there can improve a lot more.
3. I've done a little Angular JS dev but it is way outside of my comfort zone. I think the changes I've made should suffice and fit in with the existing design model but please throw any constructive criticism at me. 
4. I tested the deb on OctoPi 0.17 without issue but I am only one person so this should definitely be tested thoroughly.
5. **There is one issue that will need to be considered prior to merging/releasing this.** Currently OctoDash pulls releases with the assumption that everything is deb file. With this change both deb and tar.gz files would be listed in github releases. This may cause a condition where OctoDash tries to download and install a tar.gz as if it was a deb.
6. I'm not sure what the best way to handle package-lock.json is so I simply merged in what was used for the previous commit when rebasing. One dependency I brought in is `electron-root-path`. There could be another that I'm missing. Let me know the best way to handle this. Currently I get an error about the package-lock version being different. I'm using:
```
$ node -v
v14.15.4

$ npm -v
6.14.10
```

Phew :exploding_head:


UnchartedBull:
Great thank you! :) Going to have a look at that once v2.2.0 is released. Please don't worry about the pipeline / merge conflicts so far.

pciavald:
@UnchartedBull did you mean until v3.0.0 is released? What would be the right milestone for this PR?

UnchartedBull:
This will be the main feature of v3.0.0, since this are actually breaking changes. V3.0.0 should have a rather short Changelog with this being the main feature.

kantlivelong:
Ran into a few issues on non-OctoPi installs which I'll be looking into. One was with the user/group config and another was with the service not playing nice if lightdm was already installed. Will also be looking to see if I can remove the need for `su` in the service.

UnchartedBull:
I also run a non-OctoPi image. Problem is that my V-Core 3 just got shipped so the printer is blocked for the next week or so. After that I'm going to start testing though :)

UnchartedBull:
Oh didn't knew stale bot is also going to close PRs :o. Sorry for that

Hillshum:
This is some pretty impressive work here. Unfortunately, with the packaging of OctoDash getting changed (see #4873 ) most of this is becoming irrelevant. Some of the work around systemd is perhaps still relevant, but I think that's going to require significant changes to this branch to make it usable.

I'm going to close this PR, but if you're still around and interested in the project please feel free to reach out.

PR #1780, opened 2021-05-08
update toggles and fix plugin selection

UnchartedBull:


pciavald:
@UnchartedBull what is src/assets/toggle-switch.json ?

UnchartedBull:
Lottie Animation for the new checkbox style. https://lottiefiles.com/5222-switch. Currently only used in the setup screen, but they'll be used in the settings as well (probably for v3).

PR #1785, opened 2021-05-10
Added Profile selection and Standby Clock

makers-cave:
The Pull Request contains a new profile selections screen. Users can click the printer name and available profiles, created in OctoPi will be populated. Users can click on profile and hit the select button to switch the profile. The action will disconnect the current connection and re-connect with a new profile. A check is made for an ideal printer i.e. the profile selection screen will only populate when the printer is not doing any print job.

A clock is added to the stand-by screen with some animation to save the screen.


makers-cave:
Has old code, created a new pull request with v2,2,0 as a base.

PR #1792, opened 2021-05-11
Adjust filament change messages

pciavald:
Client returns have shown that the filament change procedure could gain a little in clarity my modifying the messages. Here's our proposition of changes.

### DIFF
--- a/src/app/filament/change-filament/change-filament.component.html
+++ b/src/app/filament/change-filament/change-filament.component.html
@@ -1,6 +1,6 @@
 <ng-container *ngIf="!usingM600(); else M600Change">
   <span class="filament-heading" i18n="@@load-new-filament">load new filament</span>
-  <span class="change-filament__info" i18n="@@load-new-filament-message">only put a little filament in, I'll pull in the rest.</span>
+  <span class="change-filament__info" i18n="@@load-new-filament-message">push the filament until feeling a slight pressure and press done</span>
 </ng-container>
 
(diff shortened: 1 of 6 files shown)


src/app/filament/change-filament/change-filament.component.html
```diff
@@ -1,6 +1,6 @@
 <ng-container *ngIf="!usingM600(); else M600Change">
   <span class="filament-heading" i18n="@@load-new-filament">load new filament</span>
-  <span class="change-filament__info" i18n="@@load-new-filament-message">only put a little filament in, I'll pull in the rest.</span>
+  <span class="change-filament__info" i18n="@@load-new-filament-message">push the filament until feeling a slight pressure and press done</span>
```
UnchartedBull, line 3: this might work for your specific printer model, but isn't applicable to every printer. So i would prefer keeping the original text here or something more general.

src/app/filament/purge-filament/purge-filament.component.html
```diff
@@ -1,4 +1,4 @@
-<span class="filament-heading" i18n="@@purging-filament">purging filament</span>
+<span class="filament-heading" i18n="@@purging-filament">Check your filament color. If the color is right, click done.</span>
```
UnchartedBull, line 1: purging is the official term and is shorter plus it fits better with the other screens ('loading filament', 'unloading filament', ...)

UnchartedBull (changes requested):

pciavald:
Will keep these changes downstream :+1:

PR #1794, opened 2021-05-12
Refactor main.js, enable custom screensize

### LINKED ISSUE #1586: Configure window size and position

For our application, we need to let the top bar (lxpanel) appear above OctoDash. We do not want to hide a part of octodash and making lxpanel appear above octodash's layer is rather complex anyway.

I've implemented the settings we need in main.js on our branch, but that's just one more hard-coded config with --dev, --big and the fullscreen setting. Functionally, I think the most elegant situation would be to read OctoDash config file before creating the window to check if a custom size and position of the window is defined. If that seems like the right way for you too, i can push a PR.

pciavald:
I've extracted most of the electron-related logic into electron.js and grouped together behaviors that were a bit scattered in main.js. The logic is more state based and with the default values for BrowserWindow and state overrides, it was easy to implement custom window properties from the user config.

As an example, we use it with these settings:
```
{
  "config": {
    octodash": {
      "window": {
        "width": 800,
        "height": 454,
        "y": 26,
        "fullscreen": false
      }
    }
  }
}
```
`x` is also exposed and was not included in this example to show how only needed values can be passed.

If you prefer that these are written by default in the config file, i can do it based on what's currently happening. Feel free to close if you don't want it in upstream.

Fixes #1586
Supplements #1627


pciavald:
package-json.lock needs to be regenerated

package-lock.json
```diff
@@ -693,9 +692,9 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.13.15",
```
pciavald, line 696: needs to be regenerated on your side

src/app/config/config.default.ts
```diff
@@ -126,5 +126,13 @@ export const defaultConfig: Config = {
     previewProgressCircle: false,
     screenSleepCommand: 'xset dpms force standby',
     screenWakeupCommand: 'xset s off && xset -dpms && xset s noblank',
+    window: {
+      width: null,
+      height: null,
+      x: null,
+      y: null,
```
UnchartedBull, line 133: numbers would make more sense here i guess.

src/app/config/config.default.ts
```diff
@@ -126,5 +126,13 @@ export const defaultConfig: Config = {
...
+      fullscreen: null,
+      backgroundColor: null,
```
UnchartedBull, line 135: please use the OctoDash background-color here (or black)

helper/config.js
```diff
@@ -8,14 +8,24 @@ let store;
...
+    config = store.get('config')
+  } catch (e) {
+    console.error(e.message);
+  }
+  return config;
+}
+
+function readConfig(window) {
+  const config = fetchConfig()
+  if (config) {
```
UnchartedBull, line 26: I guess it would make sense to merge `fetchConfig` and `readConfig`.

helper/electron.js
```diff
@@ -0,0 +1,96 @@
...
+      url: null,
+      window: {
+        frame: false,
+        fullscreen: true,
+        useContentSize: true,
+        width: mainScreen.size.width,
+        height: mainScreen.size.height,
+        x: 0,
+        y: 0,
+        backgroundColor: '#353b48',
```
UnchartedBull, line 41: this part should be loaded from the config

helper/electron.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const config = fetchConfig();
+    if (config.octodash.window) {
+      replaceExistingProperties(
+        properties.window,
+        config.octodash.window,
+        ['width', 'height', 'x', 'y', 'fullscreen', 'backgroundColor'],
+      );
+    }
```
UnchartedBull, line 85: this can be removed if the config properties are required.

UnchartedBull: Just had another look here, only thing that should be done here is that `width, height, x, y` should be set to `mainScreen.size.width, mainScreen.size.height, 0, 0` if fullscreen is true. Otherwise nothing should be done here.

src/app/config/config.default.ts
UnchartedBull: defaults should be: `800, 480, 0, 0`, since that is the main resolution OctoDash is being developed for.

pciavald:
> Why was the protocol removed? Was this tested on an actual machine or just in a dev environment?

The protocol was removed because it was not used when using full path for languages instead of the initial virtual path that can't be used with i18n.

pciavald:
I am sorry to announce that I do not have the work power to maintain all our changes in compliance with upstream requirements. This feature has been published to our repo [CosmoDash](https://github.com/Cosmyx/CosmoDash).

Issue #1749, opened 2021-05-01
NOX theme Layer progess wrong screen position and icon missing


ipa64:
Hi,
In this case pictures are better than words.

It should be
https://i84.servimg.com/u/f84/20/23/21/77/captur13.png

I get this
https://i84.servimg.com/u/f84/20/23/21/77/img_2810.jpg

Thank you

ipa64:
Tested with theGarbz/Focus same behavior

ipa64:
until v2.1.1 app was layer-progress
from v2.1.2 app is height-progress

UnchartedBull:
Please add the requested information in the issue template. Otherwise no one can help you.

pciavald:
I'm getting the same issue, NOX theme maintainer knows about it (I can't remember which issue it was in) and is already fixing it. @ipa64 you can fix it yourself before by fiddling with the custom css. Something like this
```css
.height-indication {
  width: auto !important;
  position: fixed !important;
  top: 52vh !important;
  left: 59vw !important;
}
```

ipa64:
> I'm getting the same issue, NOX theme maintainer knows about it (I can't remember which issue it was in) and is already fixing it. @ipa64 you can fix it yourself before by fiddling with the custom css. Something like this
> 
> ```css
> .height-indication {
>   width: auto !important;
>   position: fixed !important;
>   top: 52vh !important;
>   left: 59vw !important;
> }
> ```

Thank you pciavald, 
I have already tried that without success (and many variants😂), for now my workaraound is to move the line with temperatures a bit higher and the fan line a bit lower.

thegarbz:
@ipa64 What screen resolution are you using? You said you had the problem with my focus theme as well, if I can replicate it somehow I'll try and sort it out. But right now I'm unable to see the problem. 
Can you also add a screenshot from my theme? It's curious that you're not only loading in the wrong position but you're not loading the icon...

ipa64:
@thegarbz, I am using official raspberry touchscreen 7", resolution 800x480.
On my first post you should find 2 print screens, maybe they was made from NOX theme(I forgot), but it’s the same behavior. 
Now I am using your focus theme, my workaround was to move a bit higher the temperature line, and move a bit lower the fan line, to make place for the layer progress line.
Here the result: https://i84.servimg.com/u/f84/20/23/21/77/e6db1310.jpg
Thank you

ipa64:
[...]

PR #1820, opened 2021-05-18
Fix for NOX theme for OctoDash v2.2.0

noxhirsch:
Fix for NOX theme for OctoDash v2.2.0
Related issue: https://github.com/UnchartedBull/OctoDash/issues/1749

### DIFF
--- a/themes/NOX/README.md
+++ b/themes/NOX/README.md
@@ -1,4 +1,4 @@
-# NOX theme V0.2.1
+# NOX theme V0.2.2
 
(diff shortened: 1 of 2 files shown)


jneilliii:
This aligned everything up for me, thanks @noxhirsch. Only issue I notice is while printing a vase mode print the x/y gets so long with 3 decimal point numbers that the `y` part gets shoved off screen.

UnchartedBull:
Wow that was a really quick! Thanks for the fix :)

UnchartedBull (approved):

Issue #1846, opened 2021-05-22
Custom action for controlling Enclosure plugin outputs is missing


haplm:
**Describe the bug**
As discussed on Discord, the problem is that setOutput() method for the Enclosure plugin has no associated custom action, hence it can't be used from the UI.

Related ticket is #1123.
 
**Expected behavior**
I could set one of the Enclosure plugin outputs using a custom action.

ab-tools:
While adding output support for the enclosure plug-in please also add support for PWM outputs of the enclosure plug-in as well (not just the regular outputs).

I've already created a pull request to add back-end support for this to the `EnclosureService` here:
https://github.com/UnchartedBull/OctoDash/pull/1848

UnchartedBull:
https://github.com/UnchartedBull/OctoDash/wiki/Custom-Actions#predefined-actions wiki has been updated as well. Will be part of next release. If you can't wait feel free to build from main once the linked PR has been merged :)

PR #1848, opened 2021-05-23
Support PWM Outputs of Enclosure Plug-In

ab-tools:
This adds back-end support for PWM outputs to the `EnclosureService`.

Please also consider enclosure plug-in PWM outputs while working on #1846.

### DIFF
--- a/src/app/services/enclosure/enclosure.octoprint.service.ts
+++ b/src/app/services/enclosure/enclosure.octoprint.service.ts
@@ -85,9 +85,9 @@ export class EnclosureOctoprintService implements EnclosureService {
       .subscribe();
   }
 
-  setPWM(identifier: number, duty_cycle: number): void {
+  setPWM(identifier: number, dutyCycle: number): void {
     const pwmBody: EnclosurePWMBody = {
-      duty_cycle,
+      dutyCycle,
     };
     this.http
       .patch(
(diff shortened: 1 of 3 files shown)


UnchartedBull (approved):
Thanks for your contribution :)

UnchartedBull:
Please make sure that your Code adheres to the style guidelines (`npm run lint`), so the pipeline also gives a green light :)

ab-tools:
The one variable should be adjusted to camel case now.

src/app/services/enclosure/enclosure.octoprint.service.ts
```diff
@@ -84,6 +85,24 @@ export class EnclosureOctoprintService implements EnclosureService {
       .subscribe();
   }
 
+  setPWM(identifier: number, dutyCycle: number): void {
+    const pwmBody: EnclosurePWMBody = {
+      dutyCycle,
```
UnchartedBull, line 90:
```suggestion
      /* eslint-disable camelcase */
      duty_cycle: dutyCycle,
```

I think Typescript doesn't automatically convert camelCase to snake_case

ab-tools: Indeed, fixed.

PR #1850, opened 2021-05-23
Update to all Garbz Themes to suit 2.2.0

thegarbz:
All themes updated to suit Layer Progress changes
Additional tweaking to vertical scroll bar height on some themes.
Removed additional by-line as it broke on some themes and isn't part of any upstream checks.


UnchartedBull (approved):
Thanks :)

PR #2066, opened 2021-07-19
Display print success

Seculo9112:
Displays print success in file view screen and thumbnail if not printed yet


SeekaMartin (approved):
Need to update classes to be linked to variables

PR #2084, opened 2021-07-23
Display print success

### LINKED ISSUE #1339: Marking of files
Hello,

it would be nice to have some kind of marking on the print files to see if they where printed.
Either the time they where printed the last time (as in Octoprint) or be just marking them be colour (e.g. white=not printed; green=successfully printed; red=print failure).

Regards

Jens

Seculo9112:
Displays print success or thumbnail in file menu

Closes #1339.


jneilliii:
@UnchartedBull sorry for the manual build action slipping in on this PR. You might find it useful, I use it to make the deb files since electron complains about packaging on my windows. It builds, packages, and uploads as an artifact to the action run.

src/app/files/files.component.html
```diff
@@ -68,7 +68,13 @@
       [matRippleUnbounded]="false"
       class="files__object"
     >
-      <img src="assets/object.svg" class="files__icon" />
+      <div class="files__icon_wrapper {{file.successfull}}">
```
UnchartedBull, line 71: Can we use `ngClass` for the `file.successfull`? https://angular.io/api/common/NgClass

src/app/files/files.component.html
```diff
@@ -68,7 +68,13 @@
...
+        <fa-icon
+          [icon]="['fas', file.successfullIcon]"
+          class="custom-actions__action-icon"
+        ></fa-icon>
+        <img src={{file.thumbnail}} class="files__icon" />
```
UnchartedBull, line 76:
```suggestion
        <img [src]="file.thumbnail" class="files__icon" />
```

src/app/files/files.component.html
```diff
@@ -78,7 +84,7 @@
           {{ file.printTime }}<span class="files__info-unit" i18n="@@files-h">h</span>
         </span>
       </div>
-      <span class="files__name">
+      <span class="files__name" >
```
UnchartedBull, line 87:
```suggestion
      <span class="files__name">
```

src/app/services/files/files.octoprint.service.ts
```diff
@@ -60,12 +60,15 @@ export class FilesOctoprintService implements FilesService {
...
-                      ),
+                        successfull: fileOrFolder.prints != null ? fileOrFolder.prints.last.success ? 'files__object_success' : 'files__object_failed' : 'files__object_unknown',
+                        successfullIcon: fileOrFolder.prints != null ? fileOrFolder.prints.last.success ? 'check-circle' : 'minus-circle' : 'circle',
+                        thumbnail: fileOrFolder.thumbnail ? this.configService.getApiURL(fileOrFolder.thumbnail, false) : 'assets/object.svg',
+                        printTime: this.conversionService.convertSecondsToHours(
+                            fileOrFolder.gcodeAnalysis.estimatedPrintTime,
+                        ),
+                        filamentWeight: this.conversionService.convertFilamentLengthToWeight(
+                            _.sumBy(_.values(fileOrFolder.gcodeAnalysis.filament), tool => tool.length),
+                        ),
```
UnchartedBull, line 71: I think this needs to be indented one tab less.

UnchartedBull (dismissed):
PR looks good, just some minor adjustments. Please also remove the `.vs` folder and all it's contents from this PR (feel free to add it to the `.gitignore` if you want to keep them locally).

Also do the included svg icons get used at all? As far as I can see they're not used anymore, but you're now using fontawesome (really nice btw.).

And I guess the last thing: If you want to create future PRs (especially for UI stuff) it would be super neat to include a screenshot if possible, as this makes reviewing a lot easier.

Once again thanks for your contribution! Once the few remarks have been addressed this can be merged and included in the next release:`)

UnchartedBull:
@jneilliii the manual build file is actually a nice addition! Thanks for including that!

jneilliii:
@UnchartedBull I think this one is good to go now with requested changes. 

image

UnchartedBull:
Ok pipeline is green now and code looks good as well, so this can be merged.

I did some experimenting and I kinda like the coloured bars on the left side (similar to how notifications look like). I don't want to hijack anything here, but I thought, that it might be a good idea to post this here for discussion. Please let me know what you think. I have the version with the bars ready on my local dev system, so there won't be any coding effort difference between the two of them.

Screenshot 2021-08-11 at 13 11 27

jneilliii:
I've experienced this issue before from an accessibility stand pint @UnchartedBull color alone is not good enough for those people that have color blindness, specifically red/green. I personally would say to stick with the icons, but this is your project and totally up to you.

Seculo9112:
I agree color blindness could be an issue. I like them both but icons would be easier to tell them apart.

Martin Dempsey

________________________________

UnchartedBull:
Yeah color blindness is a good point. The main issue i see with the icons is that they could be hard to see on smaller screens (3.5"). I think the bars would be easier in that case.

I'd like to go with the color bars for now and if any issues occur switch back to the icons (good thing, that git keeps all the history :D). Thanks for your contribution once again! This will be released with the next release.

UnchartedBull (approved):

jneilliii:
Maybe if it were a configurable option between the two? I know from the discussion on Discord, one of the users that was involved does have color blindness.

Issue #2207, opened 2021-08-28
install issues


clue3030:
This is what I get when I try to install it and then it goes back to prompt
"E: Repository 'http://raspbian.raspberrypi.org/raspbian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'"

shueja-personal:
This is a known bug with the way Raspbian handles installing things and a status change on one of the places it looks. It's not strictly an OctoDash thing. I'm testing with the workaround we've used at CustoPiZer.

jneilliii:
So I just submitted a PR that may resolve this in the install script. It's related to an upstream debian release name change or something. We figured this out with CustoPiZer creation and the fix their resolved the same error for us.

in the meantime, this command should in theory work...
`bash <(wget -qO- https://raw.githubusercontent.com/jneilliii/OctoDash/patch-1/scripts/install.sh)`

jneilliii:
duplicate of #2206

Issue #2206, opened 2021-08-28
Install issues


clue3030:
why do I get this back when I try to install octodash?
E: Repository 'http://raspbian.raspberrypi.org/raspbian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'

Will-wastelander:
This is related to APT, not OctoDash. You would get this regard of what you are trying to install via APT.

Will-wastelander:
```
pi@Biqu-BX-octopi:~ $ bash <(wget -qO- https://github.com/UnchartedBull/OctoDash/raw/main/scripts/install.sh)
Installing OctoDash v2.2.0, armv7l
Installing Dependencies ...
E: Repository 'http://raspbian.raspberrypi.org/raspbian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
pi@Biqu-BX-octopi:~ $ sudo apt update
Hit:1 http://archive.raspberrypi.org/debian buster InRelease
Get:2 http://raspbian.raspberrypi.org/raspbian buster InRelease [15.0 kB]
E: Repository 'http://raspbian.raspberrypi.org/raspbian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
N: This must be accepted explicitly before updates for this repository can be applied. See apt-secure(8) manpage for details.
Do you want to accept these changes and continue updating from this repository? [y/N] y
Get:3 http://raspbian.raspberrypi.org/raspbian buster/main armhf Packages [13.0 MB]
Get:4 http://raspbian.raspberrypi.org/raspbian buster/contrib armhf Packages [58.7 kB]
Get:5 http://raspbian.raspberrypi.org/raspbian buster/non-free armhf Packages [104 kB]
Fetched 13.2 MB in 19s (690 kB/s)
Reading package lists... Done
Building dependency tree
Reading state information... Done
140 packages can be upgraded. Run 'apt list --upgradable' to see them.
pi@Biqu-BX-octopi:~ $ bash <(wget -qO- https://github.com/UnchartedBull/OctoDash/raw/main/scripts/install.sh)
Installing OctoDash v2.2.0, armv7l
Installing Dependencies ...
140 packages can be upgraded. Run 'apt list --upgradable' to see them.
libuuid1 is already the newest version (2.33.1-0.1).
libxss1 is already the newest version (1:1.2.3-1).
libxss1 set to manually installed.
libxtst6 is already the newest version (2:1.2.3-1).
[...]

PR #2208, opened 2021-08-29
--allow-releaseinfo-change

jneilliii:
possible fix for recent install errors, #2206, #2207

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -682,8 +682,18 @@ IFS='/' read -ra version <<< "$releaseURL"
 echo "Installing OctoDash "${version[7]}, $arch""
 
 echo "Installing Dependencies ..."
-sudo apt -qq update --allow-releaseinfo-change
-sudo apt -qq install $dependencies -y
+{
+  sudo apt -qq update
+  sudo apt -qq install $dependencies -y
+} || {
+  echo ""
+  echo "Couldn't install dependenices!"
+  echo "Seems like there is something wrong with the package manager 'apt'"
+  echo ""
+  echo "If the error is similar to: 'E: Repository 'http://raspbian.raspberrypi.org/raspbian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable''"
+  echo "you can run 'sudo apt update --allow-releaseinfo-change' and then execute the OctoDash installation command again"
+  exit -1
+}
 
 if [ -d "/home/pi/OctoPrint/venv" ]; then
     DIRECTORY="/home/pi/OctoPrint/venv"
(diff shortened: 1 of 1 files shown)


UnchartedBull:
Thanks for leading all the discussions in the issues! Was kinda busy the last couple of weeks and didn't really had time to check in. 

Just one question: Does this have any side implications, i.e. are there users that might be locked onto a different release-info?

UnchartedBull:
Just searched around the internet on how to handle that. Information seems fairly sparse, but I was able to find this:

```
--allow-releaseinfo-change
  Allow the update command to continue downloading data from a repository which changed its information of the release 
  contained in the repository indicating e.g a new major release. APT will fail at the update command for such 
  repositories until the change is confirmed to ensure the user is prepared for the change. 
  See also apt-secure(8) for details on the concept and configuration.
```

So this would indicate, that this shouldn't be done by some script, that is being downloaded from the internet. I thought that maybe pinning the issue for a couple of weeks and adding an entry to the troubleshooting guide might also work?

jneilliii:
From what we've seen it appears to only error out with apt-get, and not with just apt, so maybe that is an option. only issue is that it complains about not having a good CLI implementation, but from what I've seen with the CustOpiZer fixes with this implemented there were no bad side effects. I totally get not wanting to add it, but just trying to eliminate that from your support overhead, because it will continue...

UnchartedBull:
Yeah I totally get that with the support request. Those are usually easy to handle, but I'm afraid of maybe a ticket opening up that basically says: "hey guys, you crashed my linux installation. Thank you very much."

Interesting finding with `apt` and `apt-get` though. The script is already using `apt` though, so this might not apply in this case?

Another option would be to check for that error message and directly tell the user: Hey this happened, here is how to fix that. Might need to investigate this a little.

jneilliii:
Yeah, a sane error is definitely something to consider and I get your point. Guysoft is struggling with the same thing and I believe wants to see if this is an upstream issue with the pi-gen project.

jneilliii:
from what I understand from the linked issues, some people have mentioned just running the sudo apt update command ahead of the install script is enough to allow the install script to work, so maybe catch the error and re-run is an option.

UnchartedBull:
Hi, I updated your branch with the error message (hope that is ok). It now looks like this if apt throws an error:

Screenshot 2021-09-15 at 13 30 01

This can be merged if you're also ok with the changes :)

jneilliii:
No problem on my end.

UnchartedBull (approved):

UnchartedBull (approved):

PR #2291, opened 2021-09-23
Add German Translation

jneilliii:
Have no idea if this worked or not, but here is the translation as provided in one of the discussion threads for your review.


UnchartedBull:
Thanks for keeping this updated! I extracted the necessary parts for the translation and created https://github.com/UnchartedBull/OctoDash/pull/2692 since this PR had quite a lot of conflicts. Closing this here then.

PR #2295, opened 2021-09-26
Ophom plugin added (Phillips HUE)

ricard2k:
Added support for Ophon Pluging. Allowinf to control the Power On/PowerOff of the printer via Phillips HUE bridge.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -713,7 +713,7 @@ elif [ ! -d $DIRECTORY ]; then
 fi;
 
 if [ $DIRECTORY != "-" ]; then
-  plugins=( 'OctoDash Companion' 'Display Layer Progress (mandatory)' 'Filament Manager' 'Spool Manager' 'Preheat Button' 'Enclosure' 'Print Time Genius' 'Ultimaker Format Package' 'PrusaSlicer Thumbnails' 'TPLinkSmartPlug' 'Tasmota' 'TasmotaMQTT')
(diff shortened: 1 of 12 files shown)


jneilliii:
Why is the ophum plugin now registered on the official plugin repo?

ricard2k:
Ophom plug-in is registered:

https://plugins.octoprint.org/plugins/ophom/

jneilliii:
Oh, that's weird. My search didn't find it, maybe because if the capital letter. All good, I like to see people contributing to octodash.

ricard2k:
> Oh, that's weird. My search didn't find it, maybe because if the capital letter. All good, I like to see people contributing to octodash.

Ha, ha, the name is a little weird. I think that is the acronymous for “Octoprint HUE Outlet Manager”.

UnchartedBull (commented):
This looks really good, just two things missing:

- please add the property to the config schema: https://github.com/UnchartedBull/OctoDash/blob/main/helper/config.schema.js. This will enable OctoDash to detect, whether the value is already or there or not
- add another entry here: https://github.com/UnchartedBull/OctoDash/blob/main/src/app/app.service.ts#L56 to support upgrading the config from an earlier version.

Thanks for your contribution! If you need any help please let me know :)

ricard2k:
> * please add the property to the config schema: https://github.com/UnchartedBull/OctoDash/blob/main/helper/config.schema.js. This will enable OctoDash to detect, whether the value is already or there or not
> * add another entry here: https://github.com/UnchartedBull/OctoDash/blob/main/src/app/app.service.ts#L56 to support upgrading the config from an earlier version.
> 

Done!

PR #3053, opened 2022-07-11
🧑‍💻 Enclosure Shell Scripts Support

PS1TD:
A simple addition that lets users create custom actions that would run enclosure shell scripts using its api.


jneilliii:
I think that the placeholder `[!SHELL]` should be something more specific to identify it's enclosure plugin related. Something like `[!ENC_SHELL]` might be more appropriate?

PS1TD:
Sure ill change it to that to not confuse it with potential for the plugin to run shell scripts natively

PS1TD:
The wiki would need 1 line extra in Custom-Actions.md
`-   `"command": "[!ENC_SHELL]<identifier>"` - Runs a shell script via the Enclosure Plugin.`

UnchartedBull (commented):
Just to minor changes then this can be merged. Thanks for your contribution!

PS1TD:
Should be good now

PR #3184, opened 2022-09-14
Wemo plugin support

ssgelm:
Similar to the other PSU control plugins, this adds support for the wemoswitch plugin (https://github.com/jneilliii/OctoPrint-WemoSwitch). The plugin requires the IP and port of the wemo switch you want to control and works exactly the same as the other PSU control plugins.

I kept the locale update in a separate commit because it seems to impact more than just my changes, so I wanted it to be easy to revert if desired.


smeisner:
I was able to achieve Wemo smart switch control by:
- Adding `PSU Control` plugin to OctoPrint
- Adding `PSU Control - Wemo` plugin to OctoPrint
- Then in OctoDash Plugins, add `PSU Control` 

I now have control of the Wemo switch thru OctoDash.

jneilliii:
Yeah, that works fine if you only have one wemo to control as PSU control can only handle one. The full wemo plugin can control as many as you like.

ssgelm:
I actually wasn't aware PSUControl even had a wemo plugin 😀. Having said that I still think this has value!

jneilliii:
> I actually wasn't aware PSUControl even had a wemo plugin 😀. Having said that I still think this has value!

Yeah, I made that one too...lol.

gheydon:
Just wondering what is happening with this. I have a wemo switch on my 3d printer and would love to see this plugin merged.

PR #4152, opened 2024-02-20
Use $HOME rather than /home/pi

zandr:
`install.sh` assumes that the old default username of `pi` is in use. Now that OctoPi is in the Raspberry Pi Imager, it's fairly likely that the username will be customized, and `install.sh` will prompt with "Neither /home/pi/OctoPrint/venv nor /home/pi/oprint can be found." 

Using `$HOME` instead of hard-coding `/home/pi` fixes this.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -695,12 +695,12 @@ echo "Installing Dependencies ..."
   exit -1
 }
 
-if [ -d "/home/pi/OctoPrint/venv" ]; then
-    DIRECTORY="/home/pi/OctoPrint/venv"
-elif [ -d "/home/pi/oprint" ]; then
-    DIRECTORY="/home/pi/oprint"
+if [ -d "$HOME/OctoPrint/venv" ]; then
+    DIRECTORY="$HOME/OctoPrint/venv"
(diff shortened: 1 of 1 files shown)


metril:
There's an issue with the changes to the script. Though you're using the $HOME for all the hardcoded paths, for the changes to the sudoers file, you're still using "pi". In the context of changing to using $HOME, this is wrong.

scripts/install.sh
```diff
@@ -834,7 +834,7 @@ EOF
     sudo chmod +x ~/scripts/update-octodash
 
     sudo bash -c 'cat >> /etc/sudoers.d/update-octodash' <<EOF
-pi ALL=NOPASSWD: /home/pi/scripts/update-octodash
+pi ALL=NOPASSWD: $HOME/scripts/update-octodash
```
queengooborg, line 837:
```suggestion
$USER ALL=NOPASSWD: $HOME/scripts/update-octodash
```

queengooborg (approved):

queengooborg:
Thank you both!

Due to an issue with the CI actions that was recently resolved, this PR needs a quick sync with upstream.  Could you merge the latest upstream branch real quick?

Issue #4128, opened 2024-02-05
[Feature Request] Display IP of octoprint.


JamesM92:
***Is your feature request related to a problem? Please describe.***
-I have a octopi at my employer with octodash, but if the IP address changes there is nothing on the dash to tell me what the IP address is, and octodash took over the screen quickly making it hard to catch the IP address. 

***Describe the solution you'd like***
-Give a location in the settings menu to show what the current IP address of the octopi is encase the IP has changed regardless of any changes.

lathemaster:
I would like this option as well - display the IP address bottom center between Machine Name and Status

driftfox:
I was looking for this option in OctoDash today and realized it wasn't in there. I just created a PR that will hopefully be approved to add this to the About screen.

Sample screenshot:
![image](https://github.com/user-attachments/assets/fc5ecef6-4904-4c5b-8b08-c3148fec33d6)

queengooborg:
This has been fixed by #4633!

Issue #1320, opened 2021-01-20
IP address on main screen


scottamoulton:
I would like to request to add the ip address at the open blank space in the middle of the screen on the display.

UnchartedBull:
duplicate of #1202. please see linked issue for more details. The blank space in the middle is used for ambient temperature (if configured)

Issue #1202, opened 2020-11-20
Show IP on dashboard


dirtyheads:
**Describe the solution you'd like**
It would be nice to have the Pi's IP address to appear on the dashboard somewhere.

UnchartedBull:
will be done with #1148.

PR #4633, opened 2024-12-15
Add local IP address to About screen: Resolves #622, #1202, #1320, #4128

### LINKED ISSUE #622: Display network status icon with a current IP address
With multiple printers moving between locations I struggle figuring out what IP each printer has and if it's connected at all

A network icon indicating connection on the dashboard would be amazing with a modal info screen with IP or potentially other network details available on tap

driftfox:
I was looking for this option in OctoDash today and realized it wasn't in there. Hope you don't mind me adding it! Let me know if you have any suggestions.

_Side note: I had a difficult time getting the project to run locally. I kept receiving errors about CommonJS and Electron. On my local machine, I just went ahead and converted everything to use modules instead in order to work around this issue. I'm not sure if Dependabot just updated things to the point where CommonJS isn't supported anymore or what 🤷‍♂️_

### DIFF
--- a/src/app/model/index.ts
+++ b/src/app/model/index.ts
@@ -7,3 +7,4 @@ export * from './job.model';
 export * from './printer-profile.model';
 export * from './printer.model';
 export * from './system.model';
+export * from './util-test.model';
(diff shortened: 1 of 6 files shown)


queengooborg:
Regarding the ESM/CJS issue, this was something that I had run into as well.  I found that the `electron-store` dependency converted to pure ESM in v9, and the dependency got updated before the codebase was ready.

Since many other dependencies are going that way as well, if you're up for it, it'd be great to see a pull request for converting the project to ESM!

PR #4768, opened 2025-03-14
Upgrade (and downgrade) dependencies

queengooborg:
This PR upgrades Angular from v13 to v17, as well as upgrades many other packages, including Prettier, ESLint (with flat config migration) and more. 
 Upgrading to at least Angular v15 has already provided a huge performance benefit from what I have noticed, providing snappier page transitions.

Due to a requirement for ESM, this downgrades the prematurely-upgraded `electron-store`.  Additionally, Electron v27+ has rendering issues on the Raspberry Pi, so until that can be figured out, we should stick with v26.

Notes:
- Angular v18+ has issues where the page will be blank after the initialization screen
- Electron v27+ has rendering artifacts on the Raspberry Pi
- `electron-store` v9+ requires ESM


Hillshum (approved):

Hillshum:
@queengooborg looks good, but I’m unable to merge with the failing check

PR #4782, opened 2025-03-17
Add an option to reset the configuration

queengooborg:
This PR adds an option to reset the configuration of OctoDash, so that a user may start from scratch if they so choose.

### DIFF
--- a/helper/config.js
+++ b/helper/config.js
@@ -20,6 +20,15 @@ function readConfig(window) {
   }
 }
 
+function resetConfig(window) {
+  try {
+    store.delete('config');
+    window.webContents.send('configErased');
+  } catch {
+    window.webContents.send('configError', "Can't reset config file.");
+  }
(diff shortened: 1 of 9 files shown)


helper/config.js
```diff
@@ -20,6 +20,15 @@ function readConfig(window) {
   }
 }
 
+function resetConfig(window) {
+  try {
+    store.delete('config');
+    window.webContents.send('configErased');
+  } catch {
```
Hillshum, line 27: Does that throw any useful error info?

Hillshum (changes requested):

helper/config.js
queengooborg: Truthfully, we should never actually get to that `catch`, it's just in place in case something is seriously wrong!

Hillshum (approved):

PR #3375, opened 2022-12-31
add fan speed from companion plugin websocket messages

jneilliii:
This change will process OctoDash Companion plugin message for fan speed similar to how DisplayLayerProgress plugin was sending them. The message has been in the plugin for a while, just got around to integrating it on the OctoDash side.

PR #4784, opened 2025-03-17
Improve install and update scripts

queengooborg:
This PR improves the install and update scripts.  This includes changes from both #4152 and https://github.com/UnchartedBull/OctoDash/pull/3375/commits/90e80ab17754fd55a873c38ed1a45623008518cf.


queengooborg:
Added links to the others' changes, and added them as co-authors in the merge commit message!

PR #4785, opened 2025-03-17
Fix version comparison

queengooborg:
This PR fixes version comparison when checking if OctoPrint is at the minimum version.  This also adds better UI indications when an upgrade is required.

### DIFF
--- a/helper/discover.js
+++ b/helper/discover.js
@@ -1,12 +1,19 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-disable import/no-commonjs */
 
-const compareVersions = require('compare-versions');
+const { compare } = require('compare-versions');
 const exec = require('child_process').exec;
 
 const minimumVersion = '1.3.5';
 let browser;
 let nodes = [];
+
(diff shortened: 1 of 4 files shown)


helper/discover.js
```diff
@@ -1,13 +1,20 @@
...
+const { compare } = require('compare-versions');
 const exec = require('child_process').exec;
 
 const minimumVersion = '1.3.5';
 let browser;
 let nodes = [];
 
+function compareVersions(left, right, direction) {
+  function fixVersion(v) {
+    return v.replace('rc', '-rc').replace('--rc', '-rc');
```
Hillshum, line 13: Have you tested with some of the other version tags OctoPrint will create if running out of a git branch? I don't think thats something that needs to be supported now but I'll probably want support for that eventually.

Hillshum (approved):

helper/discover.js
queengooborg: I have not, but with my own experience and knowledge of how `git describe` works, it should work in theory!

PR #4789, opened 2025-03-17
Get fan speed from OctoDash Companion plugin

queengooborg:
This PR adds fan speed reporting via integration with the OctoDash Companion plugin.

This PR supersedes and closes #3375.


src/app/services/socket/socket.octoprint.service.ts
```diff
@@ -227,6 +233,11 @@ export class OctoPrintSocketService implements SocketService {
         set: Math.round(message?.current?.temps[0]?.bed?.target),
         unit: '°C',
       };
+      this.printerStatus.chamber = {
```
Hillshum, line 236: I'm fine with this getting slipped into this PR, but it should be mentioned in the PR text.

Hillshum (changes requested):
Did this use code from the original PR? If so I think we should credit the original author more explicitly

queengooborg:
> I think we should credit the original author more explicitly

I plan to do so by adding the original author as a co-author when performing the merge commit!

Hillshum (approved):

Hillshum:
...and here I go merging without the attribution I wanted 🙃

PR #4832, opened 2025-03-20
Include the new helpers dir in packaged build

Hillshum:
Without this packaged builds were broken


queengooborg (approved):
Oops, good catch!

PR #4861, opened 2025-03-21
Improve update flow and docs

### LINKED ISSUE #4856: OctoDash tells OctoPrint to reboot on update
The update process prompts the user on whether to reboot.  However, to do so, it sends a command to OctoPrint, which is not what it is supposed to do.  Instead, it should restart OctoDash, preferably without restarting the whole device.

### LINKED ISSUE #4860: Update instructions don't exist in README.md
There's "Update" in the table of contents but it doesn't go anywhere

Perhaps something along the lines of:
dpkg --print-architecture
wget update.deb
sudo dpkg -i update.deb
sudo service getty@tty1 restart

queengooborg:
This PR performs a few updates to the update procedure and documentation:

- Adds a section about updating to the readme (fixes #4860)
- Restarts _OctoDash,_ not OctoPrint's host system, after updates (fixes #4856)
- Allows a user to update OctoDash even if no update is available

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,16 @@ No Keyboard? [No Problem](https://github.com/UnchartedBull/OctoDash/wiki/Setup-&
 
 Having issues during the installation? Please have a look at the [Troubleshooting Guide](https://github.com/UnchartedBull/OctoDash/wiki/Troubleshooting) first.
 
+## Update
+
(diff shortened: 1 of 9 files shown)


Hillshum (approved):
Looks good

PR #4871, opened 2025-03-21
Move default config values into Ajv schema

queengooborg:
This PR moves the default values into the JSON schema, so that contributors adding plugin integration would have to modify one less file.


Hillshum:
Any way the type building can just be done at runtime?

queengooborg:
Unfortunately, no, because it's generating typedefs for TypeScript, and TypeScript is a compiled/transpiled language!

Hillshum:
Oh right. I guess I meant "can the generation be deeply integrated into the development experience so changes get auto loaded?"

queengooborg:
We might be able to use `nodemon` to monitor changes to the schema files and re-run the preparation script!

src/app/config/setup/discover-octoprint/discover-octoprint.component.html
```diff
@@ -56,14 +56,3 @@
     search again
   </div>
 </div>
-
-<div class="confirm__wrapper" *ngIf="urlSplit">
```
Hillshum, line 60: Was all this unused?

package.json
```diff
@@ -55,6 +55,7 @@
     }
   },
   "scripts": {
+    "prepare": "json2ts ./src/helper/config.schema.json > ./src/app/config/config.model.ts; node -e 'const {writeDefaultConfig} = await import(`./src/helper/config.js`); writeDefaultConfig()'",
     "build:prod": "ng build --configuration production",
```
Hillshum, line 59: Probably best to have the prod build scripts include a config regen just to be safe? Also maybe have the CI checks also verify it's been updated?

CONTRIBUTING.md
```diff
@@ -19,6 +19,9 @@ To setup a local environment, perform the following steps:
 - Install dependencies: `npm install`
 - Run the build `npm run build` (only required once to copy assets)
 
+> [!NOTE]
+> If you modify `/src/helper/config.schema.json`, run `npm run prepare` to regenerate the dependent files.
```
Hillshum, line 23: Should also indicate that this needs done when first getting started on a fresh repo too

Hillshum (changes requested):
I think this is a good step, though mostly some developer experience things maybe are missing.

CONTRIBUTING.md
queengooborg: The `prepare` command is a special command in NPM that is run when `npm i` is run, so there's no need to specifically run it on fist setup!

src/app/config/setup/discover-octoprint/discover-octoprint.component.html
queengooborg: More or less, yes -- I had never seen it pop up on my own setup!

package.json
queengooborg: The files that are generated are ignored by Git, so the CI does not need to verify they're updated since they're local only.

Regenerating during development, however, sounds like a good idea!

Hillshum (approved):

PR #4900, opened 2025-03-28
Upgrade FontAwesome to v6

queengooborg:
This PR upgrades FontAwesome to v6.

> [!NOTE]
> Some of the icons used for the default custom actions had been renamed in FA v6.


jneilliii:
Does this include the previous version shim for backwards compatibility? Otherwise it will break all custom actions users have setup using the plugin I suspect.

queengooborg:
There isn't a version shim for v5 -> v6, so a converter would have to be written.  Sadly, we wouldn't be able to catch all of the use cases.

queengooborg:
This PR upgraded us from v5, not v4.  However, it looks like the shim OctoPrint includes is for v4 -> v6, so it _might_ work...?

queengooborg:
Actually, after looking into this further, it appears that the package used for the icons is already performing shimming for us, so we're all good to go!

jneilliii:
Cool, just didn't want to have a bunch of people come back and complain because all their icons broke.

PR #4927, opened 2025-04-03
Plugin migration

Hillshum:


scripts/install.sh
```diff
@@ -771,8 +724,7 @@ xset s off
 xset s noblank
 xset -dpms
 
-ratpoison&
-octodash --no-sandbox
+chromium-browser http://localhost:5000  --kiosk
```
jneilliii, line 727: Might want to expand the options here. Taken from the official pi tutorial: https://www.raspberrypi.com/tutorials/how-to-use-a-raspberry-pi-in-kiosk-mode/
```
 --kiosk --noerrdialogs --disable-infobars --no-first-run --ozone-platform=wayland --enable-features=OverlayScrollbar --start-maximized
```

Hillshum: Hmm well the `--ozone-platform=wayland` part doesn't work without wayland (although I might move that direction anyway) but the rest is probably useful. Seems I also need to manually set the resolution (although maybe that won't be necessary on the typical size screen?) so I'll figure that out.

PR #5283, opened 2025-08-30
Send gcode from custom actions in batches

### LINKED ISSUE #4289: GCode inconsistently processed out of order
**Describe the bug**

GCode in customs actions are not always processed in order.

**To Reproduce**

1. Create a custom action with multiple GCode commands
2. Save the custom action
3. Execute the custom action

When the button for the custom action is pressed, the GCode is sometimes sent out of order.

Generally caused when using different custom actions in different orders.

**Expected behavior**

For the GCode to be processed in the order it is typed into the custom action.

**Screenshots**

I have a Prusa XL 5 tool head. I have custom actions created to pick up tool X and bring it to the front center. This allows me to inspect the nozzle and clean it if required, etc.

I also have custom actions to park the current tool and also one to bring the current tool to the front center.

If I run a **Park**, then a **Grab Tool 1**, then a **Grab Tool 3**, etc; it will inconsistently process the commands out of order.

For example:

-  the move to the front center, then get the tool, then home XY

then

-  home XY, then get the tool and then move to the center

image

image

image

image

image

This is an example of the terminal output from hitting **Grab Tool 1**, which result in the tool pickup and move being out of order.

image

then hitting **Grab Tool 3**, but this time the tool pick-up and move happen in the correct order

image

then hitting **Grab Tool 5**, and again it does it out of order

image

then telling it to **Grab Tool 2** and it does it in a completely different order

image

It's so random.

**General Information:**
 - RPi 4 4GB
 - Brand new install of OctoPi 1.0.0
 
image

 - OctoDash Version **2.3.1**
 - OctoPrint Version **1.10.1**

Hillshum:
Fixes #4289

Each custom action can contain multiple commands, separated by `;`. When multiple gcode commands are present, each is sent as a separate API call to OctoPrint. This can result in OctoPrint processing those commands out of order. 

This change remedies that by sending all gcode in a custom action as a single request to OctoPrint, guaranteeing that those will be executed in order. This change also processes any special commands before sending the gcode, although I expect that will not cause any significant issues as commands are always dispatched without waiting for results, so the exact order of execution has always been undefined and variable.

### DIFF
--- a/src/components/action-center/action-center.component.ts
+++ b/src/components/action-center/action-center.component.ts
@@ -119,15 +119,22 @@ export class ActionCenterComponent implements OnInit, OnDestroy {
         exit,
       };
     } else {
-      const parsed = this.parseCommand(command);
-      parsed.specials.forEach(action => {
-        this.executeSpecialCommand(action);
-      });
-      this.executeGCode(parsed.gcodes.join(';'));
-      if (exit) {
-        this.router.navigate(['/main-screen']);
+      try {
+        const parsed = this.parseCommand(command);
+        parsed.specials.forEach(action => {
+          this.executeSpecialCommand(action);
+        });
+        this.executeGCode(parsed.gcodes.join(';'));
+        if (exit) {
+          this.router.navigate(['/main-screen']);
+        }
+        this.hideConfirm();
+      } catch {
+        this.notificationService.error(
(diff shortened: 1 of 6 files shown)


morphias2004:
Thanks legend!  👍

PR #5314, opened 2025-09-04
v2.6.0

Hillshum:
Create version 2.6.0


queengooborg (commented):
Saw this come out just now, thanks!

Just a head's up, though, it looks like v2.6.0 accidentally got released with v2.5.4-rc.1 as the version number!

PR #5361, opened 2025-09-27
Add the API key to thumbnail requests

Hillshum:
[Version 1.2.0](https://github.com/jneilliii/OctoPrint-PrusaSlicerThumbnails/releases/tag/1.2.0) of the Slicer Thumbnails plugin requires authentication to retrieve thumbnails over HTTP. 

This PR sends the API key when request and provide blob URLs to use for `img` elements

### DIFF
--- a/src/locale/messages.xlf
+++ b/src/locale/messages.xlf
@@ -1974,21 +1974,21 @@
         <source>Can&apos;t load file!</source>
         <context-group purpose="location">
           <context context-type="sourcefile">src/services/files/files.octoprint.service.ts</context>
-          <context context-type="linenumber">168</context>
+          <context context-type="linenumber">180</context>
         </context-group>
       </trans-unit>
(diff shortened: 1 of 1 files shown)


src/services/files/files.octoprint.service.ts
```diff
@@ -148,7 +150,17 @@ export class FilesOctoprintService implements FilesService {
         map((file: OctoprintFile): string => {
           return file.thumbnail ? this.configService.getApiURL(file.thumbnail, false) : 'assets/object.svg';
         }),
-      );
+      )
+      .pipe(mergeMap(thumbnailPipe => this.getThumbnailBlobUrl(thumbnailPipe)));
```
Hillshum, line 154:
```suggestion
      .pipe(mergeMap(thumbnailPath => this.getThumbnailBlobUrl(thumbnailPath)));
  ```

PR #5385, opened 2025-10-12
Fix hotend temp in adjustment component

### LINKED ISSUE #4949: Missing Extruder in the "adjustment tab while printing
**What doesn't work?**
I recently added Octoprint with Octodash to another printer (This is not my first or only printer with this combination installed)

**What did you already try?**
I just verified that the Extruder shows up and is adjustable when not in the process of printing.

[octoprint-systeminfo-20250428115958.zip](https://github.com/user-attachments/files/19944119/octoprint-systeminfo-20250428115958.zip)

Image

Image

**General Information:**

- Hardware [e.g. Raspberry Pi, if you have layout issues please also include your screen resolution]
- OS Info [e.g. Raspbian Buster Lite, please also indicate if you used OctoPi]
- OctoDash Version [e.g. v1.0.0]
- OctoPrint Version [e.g. v1.0.0]

**Additional context**
The problem is that everything seems to be working except the missing hotend temperature and adjustment, which are only on the "adjustment" page while printing. It works fine otherwise. I have attached my system info zip file and 2 pictures, one of the main screen while printing and the other is the "adjust" page while printing. As you can the fan speed, bed temp, and speed is present but the "hotend" is blank and I one have one extruder. Thank you in advance for any help with this issue!!! Mike

Hillshum:
This got broken by the multi-filament work.

This makes the display of the first hotend temp work. Proper support for mutliple hotends will have to come later.

Fixes #4949

### DIFF
--- a/src/components/main-screen/print-control/print-control.component.html
+++ b/src/components/main-screen/print-control/print-control.component.html
@@ -53,7 +53,7 @@
               +
             </div>
             <div class="print-control__adjust__controller-value">
-              {{ temperatureHotend }}
+              {{ temperatureHotends[0] }}
               <span class="print-control__adjust__controller-value-unit">°C</span>
             </div>
             <div


richard9999999999:
I am sorry, the fix is not OK, I see the extruder temperature now (which is a good message) but when I change the temperature and push save, nothing happens, original temperature is still used..

PR #5658, opened 2026-03-01
Allow Changing OctoPrint instance

Hillshum:
Adding an endpoint to swap out the `OCTOPRINT_URL` defined in `.bashrc`, and a custom action for calling that.

Also includes:
* Using `sudo -n` on the commands so sudo doesn't prompt for a password we're not going to provide

Perhaps future work will incorporate this into the UI more natively.


octoprint_octodash/__init__.py
```diff
@@ -64,7 +65,8 @@ def on_settings_load(self):
 
     def on_settings_save(self, data):
         for plugin_name in ALL_PLUGINS.keys():
-            del data["plugins"][plugin_name]['inUse']
+            if "plugins" in data:
```
Hillshum, line 68: Investigate this

PR #5684, opened 2026-03-19
Add OctoDash start/stop/restart to menu

Hillshum:
This is borrowed from the companion plugin, so thanks to @jneilliii.

I plan to add more options around just refreshing the page as well but that will come later.


jneilliii:
Hopefully the same things doesn't happen with your implementation that seems to happen to mine. The service restarts and OctoDash does reload, but IIRC there's some kind of memory leak or orphaned octodash instance that ends up consuming resources.

PR #5711, opened 2026-04-26
feat: display spool ID in Spoolman filament selection screen

samturner3:
Follow-up to #5528 from @Hillshum - prepends the Spoolman spool ID to the display name shown in the filament selection screen (e.g., `#42 Prusament - PLA Galaxy Black`).

This makes it easier to identify spools when they are physically labeled by their Spoolman ID, like mine are, I only have the ID written on the spool.

### Changes
- Updated `displayName` in `SpoolmanOctoprintService.convertSpoolmanSpool()` to include the spool ID prefix.

### Question
Is showing the spool ID by default acceptable, or would you prefer this to be behind a configuration flag/setting? Happy to add an option for it if needed.

### DIFF
--- a/src/services/filament/spoolman.octoprint.service.ts
+++ b/src/services/filament/spoolman.octoprint.service.ts
@@ -61,7 +61,7 @@ export class SpoolmanOctoprintService implements FilamentPluginService {
       color: `#${spool.filament.color_hex}`,
       density: spool.filament.density,
       diameter: spool.filament.diameter,
-      displayName: `${spool.filament.vendor?.name || ''} - ${spool.filament.name || ''}`,
+      displayName: `#${spool.id} ${spool.filament.vendor?.name || ''} - ${spool.filament.name || ''}`,
       id: spool.id,
       material: spool.filament.material || '',
       name: spool.filament.name,


Hillshum:
Hi, thanks for the contribution!

This is a great idea and I'm happy to bring it in. I'll just do some quick tests to ensure the layout is still good.

samturner3:
Thnaks, yes please test I did not have an oppitunity to.

Also looking at the screen it may make more sence to add a new col on the far left (before the material box) for the id? Not sure if current implementation would affect (any) sorting

Hillshum:
So this what it looks like. The space to the left of the material indicates if that material is already selected (and on which tool). Seems there's a bug with that, but regardless I think the ID should stay to the right of the material. But, I think rather than adding the ID to the `displayName` string, a new column with the ID should be added. That will help get the spacing correct.

Screenshot 2026-04-27 at 8 44 39 PM

samturner3:
Sounds good, i'll take a look at adding a new column. The spacing should be the same as between the material and `displayName`? Otherwise happy for you to implement I am busy next few days

Hillshum:
Yes, the spacing should be the same on both sides. It should be tested with IDs that are 3 characters long I think too.

Also I just merged #5712 so you'll probably want to bring those changes in.

Hillshum (changes requested):
See other comment

Hillshum:
Closing as this was implemented in #5716

PR #5821, opened 2026-06-27
Fix an issue with fetching thumbnails caused by bad merge

Hillshum:


Hillshum:
Opened against wrong base

